PR #38, opened 2014-11-19
open pdf on linux with 'xdg-open'

tekknosol:
This will enable pdf opening with linux.
I'm afraid that its kind of inconvenient - I'll try to tune it.

### DIFF
--- a/lib/latex.coffee
+++ b/lib/latex.coffee
@@ -85,6 +85,8 @@ module.exports =
           require './openers/skim-opener'
         else
           require './openers/preview-opener'
+      when 'linux'
+        require './openers/preview-opener'
 
     return new OpenerImpl() if OpenerImpl?
     console.info 'Opening PDF files is not yet supported on your platform.' unless atom.inSpecMode()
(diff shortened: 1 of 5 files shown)


steelbrain:
Why not use the cross-platform Electron's `shell.openItem` https://github.com/atom/electron/blob/master/docs/api/shell.md#shellopenitemfullpath ?

thomasjo:
@steelbrain Didn't know that existed—thanks for the heads up! It could definitely work as a stop-gap solution until something better can be implemented. Since we need support for passing line information to the launched PDF viewer, using a generic API such as `shell.openItem` will not work since there isn't a well-defined standard for passing such information.

Thanks again, I'll see if I can get a simple opener implementation cobbled together tonight based on this Electron API :bow:

thomasjo:
@tekknosol Sorry about leaving this hanging for almost two years... I'm really embarrassed.

As you might have noticed, earlier this year I rewrote the entire package in JavaScript. That means I cannot merge your PR in any meaningful way. I really want to give you the credit you deserve for implementing this feature; my question is then, do you want to rewrite this in JavaScript? An alternative approach is that I rewrite everything, post the code in a Gist, then you either apply those changes to this PR or open a new one. If you don't care either way, I'll implement it myself and leave a small attribution in the header of the source file.

I totally understand whatever you decide. Apologies once more 🙇

yitzchak:
Closing as this has been addressed by #237.

PR #39, opened 2014-12-12
Add engine customization

saidai-no:
Allow selection of standard engines and custom engine command line
Move -synctex, -file-line-error and -shell-escape flags to main latexmk flags


thomasjo:
Brilliant work, just merged this. Thanks!

Issue #37, opened 2014-11-11
outputFilePath


Sasha4ever:
MAC OS
after build
Uncaught TypeError: Cannot read property 'outputFilePath' of undefined

OtoriTakeo:
Same here:

## Stack Trace 

/Users/NAME/.atom/packages/latex/lib/latex.coffee:63

TypeError: Cannot read property 'outputFilePath' of undefined
  at /Users/NAME/.atom/packages/latex/lib/latex.coffee:63:24
  at /Users/NAME/.atom/packages/latex/lib/builders/latexmk.coffee:17:9
  at ChildProcess.exithandler (child_process.js:752:5)
  at ChildProcess.emit (events.js:110:17)
  at maybeClose (child_process.js:1013:16)
  at Process.ChildProcess._handle.onexit (child_process.js:1085:5)

My system:
**Atom Version**: 0.166.0
**System**: Mac OS X 10.10.1
**Thrown From**: [latex](https://github.com/thomasjo/atom-latex) package, v0.14.0

Neither Dragynfyre nor aclemen1 suggestions helped. TeX distribution is in its default location `/usr/texbin`, although i did not install the GUI Tools.
I recently updated Tex by using `sudo tlmgr update --all`, maybe this caused the error/bug.

hortelano:
I have the same problem on both platforms: Mac OS X 10.10.1 and Ubuntu 14.04.1 LTS. Atom has been tested with fresh installations of the system and MacTeX on Mac OS X and texlive metapackage in Ubuntu. I installed the system, then MacTeX or texlive, then Atom, then language-latex and latex packages of Atom; Ctrl + Alt + B and then  the message that appears on the photos appears
![image_mac_os_x_10 101](https://cloud.githubusercontent.com/assets/10478710/5692780/025a1e8e-9904-11e4-862f-62da1c061e0e.png)
![image_ubuntu_10 4 1_lts](https://cloud.githubusercontent.com/assets/10478710/5692781/0728cc30-9904-11e4-874f-a72559d934db.png)

:

Koltharius:
Same error 

At /home/david/.atom/packages/latex/lib/latex.coffee:63

TypeError: Cannot read property 'outputFilePath' of undefined
at /home/david/.atom/packages/latex/lib/latex.coffee:63:24
at /home/david/.atom/packages/latex/lib/builders/latexmk.coffee:17:9
at ChildProcess.exithandler (child_process.js:752:5)
[...]

PR #40, opened 2014-12-27
Add Literal Haskell support. Fix (partial?) thomasjo/atom-latex#37.

aclemen1:


thomasjo:
Merged. Thanks for helping out!

PR #44, opened 2015-01-20
Update keymaps/latex.cson to fix deprecation warning

evandromr:
Use the 'atom-text-editor' tag instead of the '.editor' class

Fix the deprecation warning on Atom 0.174.0 and API 1.0


thomasjo:
Merged. Thanks for helping out!

evandromr:
:+1:

Issue #46, opened 2015-01-21
Uncaught TypeError: old path must be a string


nscaife:
If log file exists but does not contain the string "Output written on <filename>" (for example, if a previous compile failed), LogParser->parse returns *null*, which gets appended to the project directory.  

For example:

**c:\project** becomes **c:\project\null**, which does not exist, causing an exception during the file move.

**Atom Version**: 0.174.0
**System**: Microsoft Windows 8.1 Pro
**Thrown From**: [latex](https://github.com/thomasjo/atom-latex) package, v0.15.2

### Stack Trace

Uncaught TypeError: old path must be a string

```
At fs.js:580

TypeError: old path must be a string
  at TypeError (native)
  at Object.fs.renameSync (fs.js:580:18)
  at Object.fsPlus.moveSync (c:\Users\Nolen\AppData\Local\atom\app-0.174.0\resources\app\node_modules\fs-plus\lib\fs-plus.js:221:17)
  at Object.module.exports.moveResult (c:\Users\Nolen\.atom\packages\latex\lib\latex.coffee:96:8)
  at c:\Users\Nolen\.atom\packages\latex\lib\latex.coffee:39:12
  at c:\Users\Nolen\.atom\packages\latex\lib\builders\latexmk.coffee:19:9
  at ChildProcess.exithandler (child_process.js:736:7)
  at ChildProcess.emit (events.js:110:17)
  at maybeClose (child_process.js:1009:16)
  at Process.ChildProcess._handle.onexit (child_process.js:1081:5)

```

### Commands

```
 19x -1:43.9 core:page-down (ol.tree-view.full-menu.list-tree.has-collapsable-children.focusable-panel)
     -1:36.7 editor:select-to-first-character-of-line (atom-text-editor.editor)
  2x -1:36.5 core:delete (atom-text-editor.editor)
     -1:35.3 core:save (atom-text-editor.editor)
     -1:32.1 latex:build (atom-text-editor.editor.is-focused)
     -0:32.1 tree-view:remove (ol.tree-view.full-menu.list-tree.has-collapsable-children.focusable-panel)
     -0:24.4 editor:move-to-end-of-screen-line (atom-text-editor.editor)
 22x -0:24.0 core:page-down (atom-text-editor.editor)
  6x -0:22.1 core:move-up (atom-text-editor.editor)
     -0:02.4 core:move-down (atom-text-editor.editor)
[...]

PR #47, opened 2015-01-21
Add feature to clean LaTeX project directory

nscaife:
In my LaTeX workflow, it is often useful to *clean* the project directory, removing files created as a byproduct of the make process.  I've added a `Clean` item to the LaTeX menu which will take the root file and remove files with temporary extensions.  It's configurable in `latex.cleanExtensions` if non-defaults are desired.  I didn't add a keybinding, since deletion is potentially dangerous.

This was useful for me while troubleshooting #46, where I needed to do a full clean to get the document to compile and generate a new log file.


thomasjo:
Merged in 90f3863e722da9eb3530500e1d10d62375623c59. Thanks!

PR #48, opened 2015-01-22
Add support for cross-platform and Windows PDF viewers

nscaife:
These commits add support for (and dependency on) the [pdf-view](https://atom.io/packages/pdf-view) module to allow previewing directly in Atom. If no other opener is found/configured, the built-in preview is used as a fallback.

I've also added basic support for opening SumatraPDF on Windows.  Unfortunately Acrobat puts a write lock on PDF files, preventing the LaTeX build from succeeding while the PDF is open.  Sumatra will auto-refresh the PDF view on update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Current wish list, in a semi-prioritized order.
 - [ ] BibTeX autocompletion support.
 - [x] Open PDF automatically.
  - [x] Configurable.
- - [ ] Support for other distributions, besides OS X.
+ - [x] Support for other distributions, besides OS X.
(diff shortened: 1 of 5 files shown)


nscaife:
Added pdf-view to dependencies, but looks there's no support for adding dependencies onto other Atom packages.  Instead, it now detects pdf-view and will use it if available.  Otherwise, there's a friendly error suggesting to install the package.

thomasjo:
This looks extremely promising. Merging locally and testing it right now.

thomasjo:
Merged! Thanks for helping out — finally we a better Windows story :sparkles:

PR #55, opened 2015-02-10
Command to wrap long lines

mpilman:
I often collaborate on TeX-Files. My colleagues usually write long lines of text, which has several disadvantages (sync between source and PDF is less accurate and it is just ugly).

I usually wrap the lines, but I do not want to do that manually. So I just added a command that searches for all lines longer than atom.config.get('editor.preferredLineLength') and wrap them by inserting a new line. Since a new line usually does not have any semantic properties in LaTeX (other then a whitespace), this should usually work.

### DIFF
--- a/keymaps/latex.cson
+++ b/keymaps/latex.cson
@@ -1,3 +1,4 @@
 'atom-text-editor':
   'ctrl-alt-b': 'latex:build'
   'ctrl-alt-s': 'latex:sync'
+  'ctrl-alt-w': 'latex:wrap'
--- a/lib/latex.coffee
+++ b/lib/latex.coffee
@@ -68,13 +68,14 @@ module.exports =
         extension = path.extname(filePath)
         console.info "File does not seem to be a TeX file; unsupported extension '#{extension}'"
       return false
-    # we work with cursors here. This might not be the best way of doing this,
-    # but it works.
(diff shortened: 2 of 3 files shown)


thomasjo:
Just letting you know that I'm working on merging this, but it is proving somewhat cumbersome due to the fact the work is not based on the master branch. Once I believe I've succeeded, I'll push a new branch — would be great if you could play around with that to check if everything works as expected. Then once that's done, we need to implement some specs that cover the implemented functionality. But let's hold off on that until the merge is done.

thomasjo:
Just pushed a new branch called [pull-55](https://github.com/thomasjo/atom-latex/tree/pull-55) which I believe contains all of your changes. I killed all the revert commits, and based it on the current master branch. Not tested the functionality however, so let me know if you find any problems. If you feel like resolving them, please do it based on top of this new branch.
Thanks for helping out! :+1:

thomasjo:
Unfortunately we never got around to getting this merged in, and since I've changed from CoffeeScript to ES6/7 merging at this point is not viable at this point. Hopefully I'll get time to take this and your other PRs at some point, and rewrite them in ES6 or at least use them as a starting point.

I'm truly sorry about wasting your time, and for not getting them merged before the big language rewrite :bow:

PR #65, opened 2015-03-13
Improve TeX path resolution

thomasjo:
Changes the old default behavior of only using the inherited system/process $PATH environment variable if the configured (or platform default) TeX path contained the substitution marker `$PATH`. Now it instead appends the aforementioned $PATH to the configured path. $PATH substitution is still supported for those who want or need it. However, it is no longer possible to prevent inheritance, but I don't think that will cause any adverse problems.

Would be great if folks could test this out and see if it causes any regressions. I'm especially interested in feedback from those of you who use MiKTeX. You should hopefully no longer need to specify Perl paths in `latex.texPath`.

/cc @nscaife @igor25

### DIFF
--- a/lib/builder.coffee
+++ b/lib/builder.coffee
@@ -30,7 +30,11 @@ class Builder
   defaultTexPath: ->
     switch process.platform
       when 'win32'
-        'C:\\texlive\\2014\\bin\\win32'
+        [
+          'C:\\texlive\\2014\\bin\\win32'
+          'C:\\Program Files\\MiKTeX 2.9\\miktex\\bin\\x64'
+          'C:\\Program Files (x86)\\MiKTeX 2.9\\miktex\\bin'
+        ].join(';')
       else
         '/usr/texbin'
 
--- a/lib/builder.coffee
+++ b/lib/builder.coffee
@@ -20,14 +20,19 @@ class Builder
   constructPath: ->
     texPath = atom.config.get('latex.texPath')?.trim()
     texPath = @defaultTexPath() unless texPath?.length
-    texPath = texPath.replace('$PATH', process.env[@envPathKey])
+    processPath = process.env[@envPathKey]
+
+    if match = texPath.match /^(.*)(\$PATH)(.*)$/
+      texPath = "#{match[1]}#{processPath}#{match[3]}"
+    else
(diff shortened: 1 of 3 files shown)


nscaife:
Looks like this fixes the pathing problem in my testing as long as I specify `texPath`  to point to my MiKTeX bin directory.  @thomasjo do you have any problems with me adding in the default MiKTeX paths to the default path along with TeX Live?  This way, at least it wouldn't have to be specified at all on a default install.

thomasjo:
> Looks like this fixes the pathing problem in my testing as long as I specify texPath to point to my MiKTeX bin directory

Excellent! 

> do you have any problems with me adding in the default MiKTeX paths to the default path along with TeX Live?

I don't really have a problem with this, it's just that I haven't had time to verify that MiKTeX behaves identically to TeX Live when it comes to `latexmk`. Since TeX Live works the same across platforms, it was just easier for me to only officially support it. I think we can safely add MiKTeX. I'll add it now. Would be brilliant if you can verify whether it works or not.

Thanks by the way!

nscaife:
I'm happy to test.  Solid Windows support is a big priority for me.

Here are the 32-bit and 64-bit paths (I believe the 32-bit one is correct based on @igor25's previous post):

``C:\Program Files\MiKTeX 2.9\miktex\bin\x64``
``C:\Program Files (x86)\MiKTeX 2.9\miktex\bin``

thomasjo:
> Solid Windows support is a big priority for me.

Important to me as well. Sadly AppVeyor is proving to be extremely unreliable, to the point where I will likely stop using it for CI in the very near future. That will unfortunately mean no automated CI environment for Windows. I guess I'll have to rely on manually running the specs in a VM.

nscaife:
As a side note, the blue/red status text (building/failure) in the status bar is no longer appearing for me in this branch or in master.

nscaife:
I can take a look at the AppVeyor build script if you'd like.  I'll open a bug on it and we can talk there.

thomasjo:
> As a side note, the blue/red status text (building/failure) in the status bar is no longer appearing for me in this branch or in master.

I've altered some package dependencies, so it'd be advisable to remove the `node_modules` directory and execute `apm install` from the repo.

thomasjo:
Merged.

Issue #66, opened 2015-03-13
AppVeyor builds not working


nscaife:

thomasjo:
Believe the first broken spec can be fixed by something like this

``` coffee
describe "initialize", ->
  it "initializes all properties", ->
    opener = helpers.nullOpener()
    spyOn(latex, 'resolveOpenerImplementation').andReturn(opener.constructor)
    latex.initialize()

    expect(latex.builder).toBeDefined()
    expect(latex.logger).toBeDefined()
    expect(latex.opener).toBeDefined()
```

Going to spin up my Windows VM and verify.

thomasjo:
Found the bug... forgot to remove `helpers.overridePlatform('darwin')` from one of the specs during refactoring: [latex-spec.coffee#L87](https://github.com/thomasjo/atom-latex/blob/192eab770107f5b9369fcb096fb277953d96136d/spec/latex-spec.coffee#L87)

thomasjo:
AppVeyor is such an unstable platform. Most builds fail like this:
https://ci.appveyor.com/project/thomasjo/atom-latex/build/201#L7777

It's likely a combination of Chocolatey and AppVeyor, but difficult to track down since I have no means of inspecting the build environment. Perhaps it'd be better to download the latest ZIP-ed release instead of relying on Chocolatey... :thought_balloon:

thomasjo:
Had to re-launch the build a few times to make it green. I need to find an alternative to AppVeyor.

PR #67, opened 2015-03-15
Appveyor rewrite for #66

nscaife:
I spent some time today rewriting the AppVeyor YAML to assist with #66.  More specifically, I removed the Chocolatey requirement and wrapped almost everything in retry logic.  I still ran into a couple of random AppVeyor errors where the Atom install didn't complete and/or the TeX download failed, but the retry seems to resolve the issue.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -117,17 +117,25 @@ install:
 
     Write-Host Starting Atom Install...
 
-    Start-Process $atompath -argumentlist '--silent' -wait
+    $apmcmd = Join-Path -Path $env:LOCALAPPDATA -ChildPath \atom\bin\apm.cmd
+
+    $x = 0
+
+    while (!(Test-Path $apmcmd) -and $x -lt 5) {
+
+        Write-Host Starting Atom Install: Try $x
+
+        Start-Process $atompath -argumentlist '--silent' -wait
+
+    }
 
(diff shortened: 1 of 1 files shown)


thomasjo:
Thanks for helping out with this — brilliant! Looks a lot less fragile now, but there is sadly still some AppVeyor issues. I really wish I knew what the cause of each failure was; at times it's even failing to download my pre-packaged TeX Live distribution, almost as if the build agent loses network connectivity. And, as you mentioned, sometimes the actual Atom install fails.

The free build agents provided AppVeyor are vastly underpowered compared to the Travis counterpart. Merely extracting the TeX distribution takes around two times longer, and the downloads are slower...

---
*Unless we can sort out the issues soon, I will kill off the AppVeyor CI builds, since at this point it's only noise and provides zero benefit.*

nscaife:
The AppVeyor builds don't make any sense.  On the same commit, it will work fine for a build and then fail the next time.  Before the pull request, that commit built with no issues.

I agree it's not worth the additional time and effort to get these working.

thomasjo:
Regardless of AppVeyor builds being flakey or not, I really appreciate the effort you've put into getting them working. I'll merge this PR before I kill the builds.

thomasjo:
Merged in c9caeab818386a518557573f10f502522aed1b74. Thanks!

PR #83, opened 2015-05-18
Add option for always opening result in Atom

thomasjo:
Inspired by https://github.com/daskelly/atom-latex/commit/bc1c7ef21239e97374e96146174cd7eb5b779099.

Only supports the `AtomPdfOpener` right now, but at some point it might be necessary to expand that. Also not entirely sure whether to call the option `alwaysOpenResultInAtom` or `preferOpeningResultInAtom` or some such variation thereof. Might be necessary to add a description to the option as well, stating the option will only be used if the `pdf-view` package is actually installed.

- [x] Implement option `latex.alwaysOpenResultInAtom`.
- [x] Add specs.

/cc @daskelly

### DIFF
--- a/lib/config-schema.js
+++ b/lib/config-schema.js
@@ -3,6 +3,11 @@
 import {heredoc} from "./werkzeug";
 
 export default {
+  alwaysOpenResultInAtom: {
+    type: "boolean",
+    default: false,
+  },
+
   cleanExtensions: {
     type: "array",
     items: {type: "string"},
--- a/lib/latex.js
+++ b/lib/latex.js
@@ -27,6 +27,8 @@ export default class Latex {
     defineDefaultProperty(this, "builder");
     defineDefaultProperty(this, "logger");
     defineDefaultProperty(this, "opener");
+
+    this.observeOpenerConfig();
   }
 
   getBuilder() { return this.builder; }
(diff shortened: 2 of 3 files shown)


nscaife:
What about a drop-down selector where the user explicitly selects which opener to use? It would allow the existing path fields (Skim and Sumatra, and potentially others in the future) to be condensed into a single path setting.

daskelly:
Just found this package today but that's exactly what I was thinking too. Alternatively there could be a checkbox for "Open PDF in atom (requires pdf-view package)" along with another box for the user to specify the path to the desired viewer if the atom PDF box is unchecked.

thomasjo:
> What about a drop-down selector where the user explicitly selects which opener to use?

I think I like this idea. I'll try to get this PR updated tomorrow with said change so we can test it out.

thomasjo:
@nscaife I played around with the idea of switching to an enum (aka drop-down list), but quickly realized one of the pitfalls with this approach; if the configured `openerPath` does not "match" the selected `opener`, what should we do? There's no sensible way to detect the mismatch, and problems arise when passing line sync arguments and so on.
In addition the consolidated story does not lend itself as well to default path values.

I have been toying with the idea of, at some point, exposing a simple "opener" service, and extracting the opener implementations into separate, miniature packages. Not sure if the additional complexity of such an approach is worth it though.

nscaife:
@thomasjo If each entry in the enum has a sensible default (like is done now with Skim and Sumatra), then the user shouldn't have to change the path from the default.  Updating the default text inside that path field based on the selected enum entry would look great, too.

Perhaps it's possible to put an indicator next to ``openerPath'' (in the settings GUI) that displays :x: or :white_check_mark: depending on whether or not the file listed in the field can be found?  There may come a time when some PDF opener accepts Skim-compatible command arguments, for example, and the user could simply select Skim and point to a different opener without code changes.

I like the idea of mini packages for various functions (like the linter-* packages), but since we support such few openers right now (and I don't see requests for more) it might be more trouble than it's worth.   The lack of dependency tracking between Atom packages is also problematic.

thomasjo:
> If each entry in the enum has a sensible default

The "problem" is that there's no way (unless I've missed something) to couple an enum entry with a value, so the default value for the opener path has to be stored outside of the config schema. But at the same time the custom value is defined in the config schema.

What about the following compromise; each opener still has its own path option in the schema, but have an enum for specifying which opener to use?

If the compromise seems sensible, I think it would make sense to call the enum `preferredOpener` or some such. If the opener executable path points to something non-existent, default to the platform's default opener (only OS X for since it's the only platform that ships with a sensible PDF viewer out of the box—Preview.app).

nscaife:
Interesting.  My guess was that changing the current enum value would send a callback somewhere that a default path could be injected if the current setting is empty.

I think the compromise is a good idea.  Though I think if the user has misconfigured something, it is better to post a message to the user instead of performing a different action, which might be confusing.  What do you think?

thomasjo:
Yeah, I think making less guesses/assumptions makes sense. As long as the defaults are good enough to provide the best possible out of the box experience.

My basic premise with the package is that for 90% of users, no configuration should be necessary assuming they have a supported TeX distribution installed in one of the de-facto standard locations, and similarly with a supported PDF viewer. The story is only really good on OS X, sadly. I hope to find some time to improve the Linux story somewhat.

thomasjo:
Thanks heaps for your input and thoughts by the way. Very helpful!

nscaife:
No problem!  The story on Windows is quite good, IMO, and has come a long way in the past few months.  If you use TeX Live and install Sumatra, no configuration is required whatsoever.  Sumatra also supports SyncTeX and forward/back references between the document and the source, and this works great as well.

Issue #82, opened 2015-05-18
Open log when clicking "LaTeX compilation error"


nscaife:
Is it possible to have the LaTeX log open in a new tab when clicking "LaTeX compilation error" in the status bar?  Currently it opens the Atom developer pane, which is not intuitive.  

In fact, I've caught myself several times today clicking the status bar, expecting it to open the log. :8ball:

nscaife:
Bonus points if it can jump to the error when opening the log! :100:

thomasjo:
Great idea! Opening shouldn't be hard, so I'll see if I can find time to implement that tomorrow. Jumping to the first error would be trivial if my error parsing stuff was halfway decent. I need to improve that stuff soon...

myfavouritekk:
@thomasjo Is there a way to see the actual latex compilation error message for the moment?

myfavouritekk:
@thomasjo Ok, I found the error in the `.log` files. My bad, sorry. Right, showing the error message when clicking the status bar would be really handy.

thomasjo:
Just published v0.24.0.

PR #84, opened 2015-05-21
Open log file when clicking error indicator

thomasjo:
See https://github.com/thomasjo/atom-latex/issues/82.

- [x] Open log file when clicking indicator.
- [x] Scroll to first error after opening log file.

/cc @nscaife @myfavouritekk

### DIFF
--- a/lib/composer.js
+++ b/lib/composer.js
@@ -175,7 +175,7 @@ export default class Composer {
   }
 
   showError(statusCode, result, builder) {
-    this.showErrorIndicator();
+    this.showErrorIndicator(result);
     latex.log.error(statusCode, result, builder);
   }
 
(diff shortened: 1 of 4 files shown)


nscaife:
@thomasjo First commit works for me! Log opened on click.

thomasjo:
Does it make sense to destroy the error indicator when clicked? Maybe it's valuable to just keep the current behavior of persisting until the next build? Just thinking if the user closes the log file, but would prefer to open it again without navigating to the file or running another build.

nscaife:
I can see some great uses for leaving it, but I wouldn't open the log more than once.  If it's already open, just switch to that tab.

thomasjo:
> If it's already open, just switch to that tab.

Yep, that's the current behavior :smile:

Issue #87, opened 2015-05-29
Outputs are always placed in the home root


arkjinli:
The Atom version is 0.201.0, and I am using Ubuntu. The problem is: outputs are always placed in the home root, rather than where the .tex file is

This is the first time I had this problem, just after I updated the editor and the latex package.

The console error is "TeXification failed with status code 12".

Can you help me? Thank you very much!

thomasjo:
If you could provide me with some additional information, that'd be brilliant;
  1. The output of executing `apm list --installed` from a terminal.
  2. Have you specified a value for the **Output Directory** (`latex.outputDirectory`) option?
  3. The contents of a `.log` file associated with your issue.

Thanks in advance.

arkjinli:
Hi, sorry for the late reply.

Here is the information you want:

1. The out put of executing `apm list --installed` from a terminal:
    ```
    /home/ark/.atom/packages (4)
    ├── language-latex@0.6.1
    ├── latex@0.24.0
    ├── pdf-view@0.22.0
    └── script@2.23.0
    ```
    I am a new user to Atom so I didn't install many packages.

2. Have I specified a value for the **Output Directory** option?
    **No**, I leave the output directory blank.

3. The contents of a `.log` file. I created a minimum example which is basically a `hello world!` file:
    ```latex
    \documentclass{article}

    \begin{document}
        Hello world!
    \end{document}
    ```
    The corresponding `.log` file is
    ```
    This is pdfTeX, Version 3.14159265-2.6-1.40.15 (TeX Live 2015/dev/Debian) (preloaded format=pdflatex 2015.5.29)  29 MAY 2015 23:05
    entering extended mode
     restricted \write18 enabled.
     file:line:error style messages enabled.
     %&-line parsing enabled.
    **helloworld.tex
    (./helloworld.tex
    LaTeX2e <2014/05/01>
    Babel <3.9l> and hyphenation patterns for 79 languages loaded.
    (/usr/share/texlive/texmf-dist/tex/latex/base/article.cls
    Document Class: article 2014/09/29 v1.4h Standard LaTeX document class
    (/usr/share/texlive/texmf-dist/tex/latex/base/size10.clo
    File: size10.clo 2014/09/29 v1.4h Standard LaTeX file (size option)
    )
    \c@part=\count79
    \c@section=\count80
    \c@subsection=\count81
    \c@subsubsection=\count82
    \c@paragraph=\count83
    \c@subparagraph=\count84
[...]

PR #94, opened 2015-06-18
Fix working directory bug

thomasjo:
Hopefully this resolves the bug reported in #87, but I haven't actually tested if it resolves the problem. Would be brilliant if someone affected by the bug can test this PR, and report whether it's resolved. 

Extremely busy right now, so I don't have time to test this myself until later tonight or tomorrow.

/cc @arkjinli, @Victorystick, @Lednerb


thomasjo:
Too complex, and not working as expected across platforms.

PR #95, opened 2015-06-19
Fix working directory issue (redux)

thomasjo:
Second attempt at fixing #87, this time with a simpler approach that should work as intended across platforms. Again, would be really great if people can test this to verify that it works as intended.

Will :boat: this tonight unless someone finds a bug or some such.

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -8,7 +8,7 @@ export default class Builder {
     this.envPathKey = this.getEnvironmentPathKey(process.platform);
   }
 
-  run(/* args, callback */) {}
+  run(/* filePath */) {}
   constructArgs(/* filePath */) {}
   parseLogFile(/* texFilePath */) {}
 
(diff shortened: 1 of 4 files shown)


thomasjo:
I've tested this a fair bit on Ubuntu 15.04 and I'm convinced this PR resolves #87, so unless I hear otherwise, I will merge and publish a patch release tonight.

/ping @Lednerb, @arkjinli, @Victorystick

Lednerb:
Tested it and it works like it should.
Everything's fine :+1: 

The added option for project folder selection... is this not the same as "Move Results to Source Directory" ?

thomasjo:
@Lednerb No, I decided to not ship the option previously discussed—I closed the [old PR](https://github.com/thomasjo/atom-latex/pull/94) and opened this one instead.

I might consider the feature again in the future, but it didn't work as intended across platforms with the way I had implemented it, and I'd rather focus on fixing the bug. Hopefully that makes sense :smiley:

Lednerb:
Ah okay, i've cloned the wrong branch.
`tj-fix-cwd-issue-redux` works fine.

PR #97, opened 2015-07-01
Add crtl-alt-c => clean shortcut

balu-:


nscaife:
I originally omitted this to prevent someone from accidentally cleaning, since there's no confirmation before it irrevocably deletes the files.  The defaults are fairly safe, though, and shouldn't delete something that can't be regenerated.

balu-:
Oh, I see…
just thought it was missing and tried to be a good open source citizen :wink:

nscaife:
I certainly don't see an issue with it, and it's great that you submitted a pull request (I wish more people would!).

thomasjo:
Thanks! I'm a little bit reluctant due to the same reason as @nscaife, but if it becomes a problem we can fix it then.

PR #100, opened 2015-07-08
Tweak TexifyBuilder

thomasjo:
Currently `TexifyBuilder` has some incorrect feature flags, as well as functionality that is not (properly) supported by `texify` such as engine customizations and `--output-directory`.

The problematic functionality is removed and instead replaced with notifications to let the user know they are using unsupported functionality. I wish there was a better way, but configuration is still a static concept in Atom, so for the time being this way of handling unsupported config combinations will suffice.

### DIFF
--- a/lib/builders/texify.js
+++ b/lib/builders/texify.js
@@ -48,7 +48,7 @@ export default class TexifyBuilder extends Builder {
 
     let outdir = this.getOutputDirectory(filePath);
     if (outdir) {
-      atom.notifcations.addInfo(heredoc(`
+      atom.notifications.addInfo(heredoc(`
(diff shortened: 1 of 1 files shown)


thomasjo:
I'm going to merge this in a few hours unless anyone has objections. I've tested it (manually) in a Windows 8.1 VM with MiKTeX 2.9, and everything seems to work as expected now. And the things that don't behave correctly have been disabled and the user is presented with some warning notifications.

I should probably update the (configuration wiki page)[https://github.com/thomasjo/atom-latex/wiki/Configuration] with information about the functionality that isn't supported with `texify`.

PR #114, opened 2015-08-18
Wrap custom opener file path in quotation marks to support spaces.

ahoereth:
If ${filePath} contains spaces most PDF viewers, e.g. evince, will not treat it as a single argument. Wrapping the path in quotation marks fixes that issue.


thomasjo:
@ahoereth Thank you! :bow:

PR #135, opened 2015-10-27
Fix sumatra opener on Windows 

m0nhawk:
I've also fixed the Sumatra opener, it was broken in 0.27.1

### DIFF
--- a/lib/openers/sumatra-opener.js
+++ b/lib/openers/sumatra-opener.js
@@ -5,15 +5,9 @@ import Opener from '../opener'
 
 export default class SumatraOpener extends Opener {
   open (filePath, texPath, lineNumber, callback) {
-    const sumatraPath = `"${atom.config.get('latex.sumatraPath')}"`
-    const args = [
-      '-forward-search',
-      `"${texPath}"`,
-      lineNumber,
-      `"${filePath}"`
-    ]
+    const command = `"${atom.config.get('latex.sumatraPath')}" -forward-search "${texPath}" "${lineNumber}" "${filePath}"`
 
-    child_process.execFile(sumatraPath, args, (error) => {
+    child_process.exec(command, (error) => {
       if (callback) {
         callback((error) ? error.code : 0)
       }


m0nhawk:
I would be also thankful to get an idea on how to update my fork with the upstream (`thomasjo:master`) without merge commits...

lib/openers/sumatra-opener.js
```diff
@@ -5,15 +5,9 @@ import Opener from '../opener'
...
 export default class SumatraOpener extends Opener {
   open (filePath, texPath, lineNumber, callback) {
-    const sumatraPath = `"${atom.config.get('latex.sumatraPath')}"`
-    const args = [
-      '-forward-search',
-      `"${texPath}"`,
-      lineNumber,
-      `"${filePath}"`
-    ]
+    const command = `"${atom.config.get('latex.sumatraPath')}" -forward-search "${texPath}" "${lineNumber}" "${filePath}"`
```
thomasjo: I'd really prefer to keep the `sumatraPath` and `args` variables as it makes it a little bit easier to reason about how the command is constructed. The actual `command` variable itself could then perhaps be pieced together as
```js
const command = `"${sumatraPath}" ${args.join(' ')}`
```

thomasjo:
Thanks for helping out with this. I left you a line comment that needs to be addressed.

---
As for avoiding the merge commit, the way I usually approach is through `git rebase`. Let's say I'm on a branch called `awesome-feature` and it has fallen behind `master`, then I would rebase `awesome-feature` on top of the `HEAD` of the `master` branch. Now assuming you're currently on the feature branch, simply execute `git rebase master`. This will replay all the commits in the feature branch "on top" of `master`.

In your case the workflow might be something along the lines of 
``` bash
$ git fetch thomasjo # looks like this is what your remote is called
$ git checkout <your-feature-branch> # only if you're not already on the feature branch
$ git rebase thomasjo/master # or equivalent refspec
```

Right now it looks like you've been making changes in your `master` branch, but I would highly advise against doing that. Make a habit out of making all changes in a feature branch. Simply do 
``` bash
$ git checkout -b <some-descriptive-branch-name>
```

m0nhawk:
Thanks, I've fixed this in the new branch.

I'll then go with the `fetch`-`rebase` workflow on later feature. So the `merge` won't mess up in pull requests.

PR #140, opened 2015-11-03
Fix sumatra opener on Windows

m0nhawk:
I've also fixed the Sumatra opener, it was broken in 0.28

### DIFF
--- a/lib/openers/sumatra-opener.js
+++ b/lib/openers/sumatra-opener.js
@@ -9,11 +9,13 @@ export default class SumatraOpener extends Opener {
     const args = [
       '-forward-search',
       `"${texPath}"`,
-      lineNumber,
+      `"${lineNumber}"`,
       `"${filePath}"`
     ]
 
-    child_process.execFile(sumatraPath, args, (error) => {
+    const command = `${sumatraPath} ${args}`
+
+    child_process.exec(command, (error) => {
       if (callback) {
         callback((error) ? error.code : 0)
       }


lib/openers/sumatra-opener.js
```diff
@@ -9,11 +9,13 @@ export default class SumatraOpener extends Opener {
     const args = [
       '-forward-search',
       `"${texPath}"`,
-      lineNumber,
+      `"${lineNumber}"`,
       `"${filePath}"`
     ]
 
-    child_process.execFile(sumatraPath, args, (error) => {
+    const command = `${sumatraPath} ${args}`
```
thomasjo: Unless there's something odd about how Sumatra parses arguments, this will have to be changed to 
~~~ js
const command = `${sumatraPath} ${args.join(' ')}`
~~~
otherwise the arguments will end up being comma-separated since that's the result of `Array#toString()`.
For reference see my commit https://github.com/thomasjo/atom-latex/commit/ea338aaf33218963159021f1d94b05f15631b7d8, but I wanted to give you a chance to fix this. If not I can happily merge this PR and apply my fix afterwards :bow:

thomasjo:
Thanks again! I left a line comment for you. Besides that, this looks good. I'll merge after you've responded to my comment.

lib/openers/sumatra-opener.js
m0nhawk: The fix is pretty mandatory, I just haven't tested (as this is pretty simple edit) and it will fail to open otherwise.

I don't mind mergin and applying your fix.

Issue #125, opened 2015-09-27
Building causes a new sumatra-pdf window every time


mmoh:
Every time I build a late file a new sumatra window is opened. This is quite annoying since you end up with many sumatra pdf windows open at the same time, unless you close it every time your done after a build.

thomasjo:
Thanks for the report. I agree that this is annoying, and very unfortunate. I'll see if I can figure out a sensible way of fixing this problem.

m0nhawk:
@mmoh As a workaround I can offer to set the "Use tabs" option in the Settings.

mmoh:
@m0nhawk  I can't seem to find this option? 
@thomasjo  One thing that does work is to disable the option to automatically open the pdf file, open the file manually using sumatra, and then then sumatra automatically gets refreshed  when you build. 

I'm not sure what open results in the background does but on my computer in has no effect.

m0nhawk:
@mmoh And what Sumatra version you use? I have 3.0 (latest). It's in the Advanced section.

![test](https://cloud.githubusercontent.com/assets/8220926/10140105/ab96fe72-660e-11e5-98d3-5bc4277af3f5.png)

nscaife:
I'm not able to reproduce this.  My system refreshes the PDF in the same Sumatra window.

mmoh:
@nscaife Are you on Windows?
@m0nhawk  I'm using version 3 like you. Ok thank's I see the setting now.

The sublime text latex plugin doesn't have this problem. So perhaps that could be used for inspiration to fix this issue.

nscaife:
@mmoh I am on Windows, Sumatra v3.0.

mmoh:
Ok so it looks like it works when I select  the use tabs option, which I normally don't use. Still this should be fixed so that it works in both cases. Perhaps a simple solution is on the first compilation to check if the pdf is loaded in sumatra, in subsequent compilations not to launch sumatra, since the pdf file will refresh and sumatra will detect these changes.

nscaife:
In order to do that, I think the plugin would have to either:

1. Only open on first compile, which may be problematic if someone accidentally closes Sumatra. This could probably be made a config setting, though.
2. Scan window titles looking for a Sumatra instance with a title including the PDF file name, which may be problematic if you have multiple PDFs with the same name.

Perhaps another good solution would be to file a bug on the Sumatra repository, since it may not be expected behavior.

mholzhauser:
[...]

PR #141, opened 2015-11-04
Fix Sumatra multiple instances problem reported in #125

m0nhawk:


thomasjo:
Thanks @m0nhawk for fixing this, and thanks to @mholzhauser for the info on the `-reuse-instance` argument flag!

PR #143, opened 2015-11-05
Add support for ps and dvi file formats for latexmk builder

sometea:
Hi, I have added an option to create ps or dvi files instead of the default pdf when using the latexmk builder. Currently, this only works for latexmk, since I do not have access to texify.

### DIFF
--- a/lib/builders/latexmk.js
+++ b/lib/builders/latexmk.js
@@ -41,6 +41,7 @@ export default class LatexmkBuilder extends Builder {
     const engineFromMagic = this.getLatexEngineFromMagic(filePath)
     const customEngine = atom.config.get('latex.customEngine')
     const engine = atom.config.get('latex.engine')
+    const outputFormat = atom.config.get('latex.outputFormat')
 
     if (enableShellEscape) {
       args.push('-shell-escape')
@@ -57,6 +58,13 @@ export default class LatexmkBuilder extends Builder {
     let outdir = this.getOutputDirectory(filePath)
     if (outdir) {
       args.push(`-outdir="${outdir}"`)
+    }
+
+    if (outputFormat) {
+      if (!(outputFormat=='pdf')) {
+        args.splice(args.indexOf('-pdf'), 1)
+        args.push(`-${outputFormat}`)
+      }
     }
 
     args.push(`"${filePath}"`)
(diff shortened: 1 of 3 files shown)


lib/builders/latexmk.js
```diff
@@ -59,6 +60,13 @@ export default class LatexmkBuilder extends Builder {
       args.push(`-outdir="${outdir}"`)
     }
 
+    if (outputFormat) {
+      if (outputFormat !== 'pdf') {
```
thomasjo: These two lines should probably be combined into
~~~ js
if (outputFormat && outputFormat !== 'pdf') {
~~~

Although since the config value is an enum, this should in reality always pass the first test, so in theory we should be able to remove the truthy check on `outputFormat`, hence reduce both if blocks down to
~~~ js
if (outputFormat !== 'pdf') {
~~~

lib/builders/latexmk.js
```diff
@@ -59,6 +60,13 @@ export default class LatexmkBuilder extends Builder {
...
+        args.splice(args.indexOf('-pdf'), 1)
+        args.push(`-${outputFormat}`)
+      }
+    }
+
```
thomasjo: The more I think about it, if we either move the `const args` initialization to be after we've initialized `const outputFormat` or move the initialization of the latter to before the former, then we should be able to something like this

~~~ js
const args = [
       '-interaction=nonstopmode',
       '-f',
       '-cd',
       `-${outputFormat}`,
       '-synctex=1',
       '-file-line-error'
     ]
~~~

If it's the case that we want to do something additional tweaking when using PS or DVI output, we could also remove '-pdf' from `const args` initialization, and simply append it to the array like we do with the other special case arguments.

thomasjo:
Overall this is brilliant work, and thank you for helping out! I left a few line comments that I would like to address before merging though. You should at least collapse those two `if` statements into a single one if you don't have time or interest in going the alternate route of the re-ordering of variable initialization. I'm happy to do that refactoring after merging, so just let me know you're preference.

sometea:
Thanks for your comments, and thanks for this great package. I have followed your second suggestion and just moved the initialization of const outputFormat to before the initialization of args. I have also added a default value to outputFormat in case it is not defined. (The test checking for the default command line arguments fails otherwise.)

thomasjo:
There are some minor things I'd like to tweak, but I can deal with that myself. Thanks so much for adding this @sometea. I really appreciate the help :bow:

sometea:
Thanks for merging. Let me know if something else comes up that I could possibly help with.

PR #144, opened 2015-11-05
knitr builder support

m0nhawk:


m0nhawk:
What is currently lacks and will be implemented:

[ ] Work on `.Rnw` files only without manual selection or special hotkey
[ ] Check `R` and `knitr` library is installed.

thomasjo:
Thanks for getting started on this! I've got some ideas on how to generalize the `Work on .Rnw files only without manual selection or special hotkey` task. I'll try to get time later today or over the weekend. The basic concept however is to "query" all the builders which ones can process the current file. The initial naive implementation would be to just implement extension checking (future version might also check file contents to be more robust). That way we can easily add more special builders like this in the feature and (hopefully) not have to change any code besides implementing the builder.

Since we'll have two builders responding to `.tex` since both `latexmk` and `texify` builders can handle those files, we will need to give them some special treatment where resolution is resolved via the `builder` config variable.

I'll let you know once I've implemented this stuff on master. Sound OK?

PR #147, opened 2015-11-09
Generalize the builder concept

thomasjo:
This is an attempt to generalize the builder concept which will make it easy to implement file type specific builders, which is currently needed in #144.

/cc @m0nhawk

---
#### Current tasks
  - [x] Add simple builder registry by simply loading all modules in the builders
    directory.
  - [x] Add `BuilderRegistry::getBuilder` method for querying registered builders
    for ones that can process the current file. Querying is done via abstract
    `Builder.prototype.canProcess` method.
  - [x] Resolve duplicate builder registration for `.tex` files. For regular `.tex` files we return both
    `LatexmkBuilder` and `TexifyBuilder`, but we need to choose one based on configuration.
    Whether this should be done in the `BuilderRegistry::getBuilder` method or in the `Composer` needs
    to be experimented with.
  - [x] Remove `Latex.getBuilder` and `Latex.builder`, and migrate to new codepath.

### DIFF
--- a/lib/builder-registry.js
+++ b/lib/builder-registry.js
@@ -3,41 +3,37 @@
 import fs from 'fs-plus'
 import path from 'path'
 
-let builders = []
-
-function getAllBuilders () {
-  if (builders.length === 0) {
-    const moduleDir = path.join(__dirname, 'builders')
-    const entries = fs.readdirSync(moduleDir)
-    builders = entries.map((entry) => require(path.join(moduleDir, entry)))
-  }
-
-  return builders
-}
-
-function resolveAmbigiousBuilders (builders) {
-  const names = builders.map((builder) => builder.name)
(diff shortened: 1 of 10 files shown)


m0nhawk:
Hi, what's the status of this issue? I'd like to finish the support of `knitr` in a package.

thomasjo:
@m0nhawk It's basically done, I just hit a minor "roadblock" on the fact that in order to have proper knitr support, we need to run both the `RScript` command **and** a LaTeX builder. I started thinking about ways of abstracting this generalization further to make that dead easy to implement. But I'm not entirely sure it's worth doing right now...

---
Basically what I was toying with was either implementing the notion of a build pipeline, or at the very least make it easier to "grab" a builder given a file path, then invoking that builder from within another. The latter approach is more or less ready.

thomasjo:
I would love to make this even better, but I think it's good enough to :ship:

PR #148, opened 2015-11-10
Make keymap selector more specific to help with other packages overlo…

pgbross:
…ading the same key sequences


thomasjo:
Thanks! Been thinking about this for a long time just never got around to it :smiley:

PR #154, opened 2015-11-28
Remove "LaTeX Error" from error message regex.

aj-foster:
[This pull request is mostly for discussion purposes, and you're obviously welcome to merge if it turns out to be a good idea.]

Many error-related output messages from pdflatex do not include this marker. By removing it, the console logger correctly displays a number of additional error messages.

Please be aware that I have not had the opportunity to test this in different environment and with different compilers. It very well may be a change that you've considered but rejected for specific reasons.

My environment:
* Atom 1.2.4
* atom-latex 0.28.2
* Mac OS X 10.11.1
* pdflatex 3.14159265-2.6-1.40.16 (TeX Live 2015)
* latexmk 4.43a

Thank you so much for you work on this; it's a joy to use!

### DIFF
--- a/lib/parsers/log-parser.js
+++ b/lib/parsers/log-parser.js
@@ -13,7 +13,6 @@ const outputPattern = new RegExp('' +
 const errorPattern = new RegExp('' +
   '^(.*):' +               // File path.
   '(\\d+):' +              // Line number.
-  '\\sLaTeX\\sError:\\s' + // Marker.
   '(.*)\\.$'               // Error message.
 )
 


thedataking:
It is important not to make the error regex too generic. Here is what a logfile for a clean build contains

```
 file:line:error style messages enabled.
Package: infwarerr 2010/04/08 v1.3 Providing info/warning/error messages (HO)
```

Just my two cents,
Per

thomasjo:
@aj-foster Thanks for taking the time to contribute! As alluded to by @thedataking, your change triggers a lot of false positives unfortunately. It catches warnings, as well as things that are purely informational.

I want to improve the log parsing routine, but unfortunately LaTeX logs are not structured—they are a royal mess—so it will require some careful considerations. In particular, warnings (and info) should probably be caught first to remove or reduce the chance of false positives, then a "looser" pattern (or set of patterns) can be used to catch errors. The thing is that various packages and backends produce their own non-standard warnings and error messages, and we need to handle all those scenarios somehow.

The current pattern is used since it doesn't produce any false positives, but it also misses a lot errors unfortunately, as you've discovered.

aj-foster:
Is there (or could there be) some sort of repository for LaTeX log snippets that could act as a corpus for future development? My environment and common errors create a (probably small) subset of errors and messages to be filtered.

Alternatively, would you be interested in an [outside solution](https://github.com/jpallen/latex-log-parser) if it is/becomes well-suited for the job?

Let me know if you'd like to close this and move discussion to an issue instead.

thomasjo:
@aj-foster I'm not aware of any such repository, but that's a great idea! Maybe we should create a wiki page in this project, alternatively a Markdown document or some such.

A third-party solution is definitely acceptable as long as the overhead is minimal and it integrates nicely. Since it's not likely to catch every error or warning, it would be great if the solution was extensible in some manner.

thomasjo:
I'm going to close this pull request for the reasons stated earlier, but thanks again for taking the time to contribute! :bow:

PR #156, opened 2015-12-11
Clean uses now custom build directory

Poeschl:
The clean command now take the custom build directory into account.


thomasjo:
Thanks for this! :bow:

PR #163, opened 2016-02-03
Added inverse-search functionality for sumatra-opener

avonmoll:
* Simply adds `-inverse-search` flag to the SumatraPDF command that `sumatra-opener.js` generates
* Inverse search means double clicking inside the generated SumatraPDF document generates a command to open the source document in Atom at the source line that produced it
* __Tested on Windows only__ -- based the command line statement found at the bottom of [this discussion](https://discuss.atom.io/t/inverse-search-with-sumatrapdf/20076/2)
* Had to add a setting for the path to the Atom executable in order to complete the arguments to the `-inverse-search` flag

### DIFF
--- a/lib/config-schema.js
+++ b/lib/config-schema.js
@@ -9,11 +9,6 @@ export default {
     default: false
   },
 
-  atomPath: {
-    description: 'Set the Path to Atom Executable (used for inverse search from SumatraPDF).',
-    type: 'string',
-  },
-
   builder: {
     description: 'Select LaTeX builder. MiKTeX distribution is required for texify.',
     type: 'string',
--- a/lib/config-schema.js
+++ b/lib/config-schema.js
@@ -9,6 +9,11 @@ export default {
     default: false
   },
 
+  atomPath: {
+    description: 'Set the Path to Atom Executable (used for inverse search from SumatraPDF).',
+    type: 'string',
+  },
+
   builder: {
     description: 'Select LaTeX builder. MiKTeX distribution is required for texify.',
     type: 'string',
--- a/lib/openers/sumatra-opener.js
+++ b/lib/openers/sumatra-opener.js
@@ -6,7 +6,7 @@ import Opener from '../opener'
 export default class SumatraOpener extends Opener {
   open (filePath, texPath, lineNumber, callback) {
     const sumatraPath = `"${atom.config.get('latex.sumatraPath')}"`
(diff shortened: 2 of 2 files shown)


thomasjo:
First of all, sorry about the delayed review! And thanks for contributing, greatly appreciated :bow: 

I'm not loving the `atomPath` option since it's a hack only necessary for this functionality to work. And also something that is likely to go stale. We will have to find a better solution for the problem at hand.
I'm not entirely sure what that solution is, but we have to find it.

thomasjo:
My initial gut feeling to avoid the `atomPath` scenario is to create some utility/helper based on processing `process.argv`, where the first element is what we want. Combining that with `process.platform` should allow us to create a platform-agnostic solution that does not require user intervention in any manner, and will just "magically" work.

What do you think? I'm happy to implement this utility for you if you want (including the necessary tests), and if you agree of course.

avonmoll:
Sorry for the delayed response - I submitted this pull request days before my son was born!

It sounds like you have a good sense for making this idea work without having to introduce an additional parameter in the Settings. I'd like to help out if possible, but I may need some more elaboration (I don't really know any JS - I hacked this small change by trial and error).

thomasjo:
> Sorry for the delayed response - I submitted this pull request days before my son was born!

Congratulations! :baby: 

I'll see if I can find time over the next few days to implement a simple helper function for getting the path to the Atom executable.

avonmoll:
Just did a bit of poking around and discovered how to access/use process.argv. The first argument is indeed the full path to the Atom executable. This seems to be all we need - unless I'm missing something?

thomasjo:
I'm going to take this for a spin on a Windows 10 VM now — I'll merge unless I find any issues.

thomasjo:
I'm experiencing some problems with this functionality; when first initiating a sync **from** SumatraPDF, it opens a new instance of Atom instead of reusing the existing one. This is likely caused by the fact I'm testing in dev mode (`atom --dev`), and thus with the `process.argv[0]` approach and no launch flags, Atom will spin up a new instance.

I'll see if I can repro from a normal Atom instance.

thomasjo:
@avonmoll Thanks again for helping out with this! I'm going to :ship: this right away.

avonmoll:
Glad I could help in a small way! Also excited to be counted among the list of contributors; this was my first Pull Request!

Issue #166, opened 2016-02-23
Log parser not parsing errors


Focus:
When I get errors in my .log file, they don't contain the term `Latex Error`. For example,
```
./coalescent_notes.tex:116: Undefined control sequence.
```
These are not picked up by the log parser and the tests are failing. My details:
OS: OSX
Latexmk version: 4.39
Latex distribution: TeX Live 2014

thomasjo:
Fixed in #167

Issue #165, opened 2016-02-17
Mark editor with warnings and errors


Focus:
Nice work on the package! I would very much like it if the lines with errors and warning were shown on the editor as a marker with a decoration (perhaps red for error and yellow for warning). Would it be possible to do this?

I have found the relevant documentation here:
https://atom.io/docs/api/v1.5.3/Decoration
https://atom.io/docs/api/v1.5.3/TextEditor#instance-markBufferRange

thomasjo:
Fixed in #167

PR #167, opened 2016-02-24
Marker implementation

Focus:
Implemented markers #165. Along the way there was an issue with the way log parser was parsing #166 which is fixed.

### DIFF
--- a/lib/error-marker.js
+++ b/lib/error-marker.js
@@ -1,7 +1,7 @@
 'use babel'
 
 export default class ErrorMarker {
-  constructor(editor, errors, warnings) {
+  constructor (editor, errors, warnings) {
     this.editor = editor
     this.errors = errors
     this.warnings = warnings
@@ -9,24 +9,24 @@ export default class ErrorMarker {
     this.mark()
   }
 
-  mark() {
-    for (error of this.errors){
+  mark () {
+    for (let error of this.errors) {
       this.markRow(error.lineNumber - 1, 'line-red')
     }
-    for (warning of this.warnings){
-      this.markRow(error.lineNumber - 1, 'line-yellow')
+    for (let warning of this.warnings) {
+      this.markRow(warning.lineNumber - 1, 'line-yellow')
     }
   }
 
-  markRow(row, colour) {
-    column = this.editor.buffer.lineLengthForRow(row)
-    marker = this.editor.markBufferRange([[row, 0], [row, column]], {invalidate: 'touch'})
-    decoration = this.editor.decorateMarker(marker, {type: 'line-number', class: colour})
+  markRow (row, colour) {
+    const column = this.editor.buffer.lineLengthForRow(row)
+    const marker = this.editor.markBufferRange([[row, 0], [row, column]], {invalidate: 'touch'})
+    this.editor.decorateMarker(marker, {type: 'line-number', class: colour})
     this.markers.push(marker)
   }
 
-  clear() {
-    for (marker of this.markers){
+  clear () {
+    for (let marker of this.markers) {
       marker.destroy()
     }
   }
--- a/lib/error-marker.js
+++ b/lib/error-marker.js
@@ -1,0 +1,33 @@
+'use babel'
+
+export default class ErrorMarker {
+  constructor(editor, errors, warnings) {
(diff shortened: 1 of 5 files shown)


thomasjo:
Thanks for the contribution! :bow: 

I've started reviewing your changes, and overall it looks awesome. There are some style problems that will have to be addressed however. For some reason unknown to me right now, the CI linting is not behaving as expected and is currently producing false positives. In the meantime to uncover said style issues you can locally execute `standard`, e.g.
```bash
$ ./node_modules/.bin/standard lib/**/*.js
```

For your convenience, I dumped the errors into a gist:
https://gist.github.com/thomasjo/c7b77bb34fa0fdcd0d8f

---
*Side note: I'm looking into the false positives right now.*

thomasjo:
The CI process now correctly fails on the style issues mentioned earlier.

---
To make it even easier to uncover these during local development, I'll look into providing some utility functions, as well as creating a contribution guide.

Focus:
Thanks for the feedback. I fixed the issues with formatting. Let me know if it needs any more changes.

lib/error-marker.js
```diff
@@ -0,0 +1,33 @@
...
+    this.editor = editor
+    this.errors = errors
+    this.warnings = warnings
+    this.markers = []
+    this.mark()
+  }
+
+  mark () {
+    for (let error of this.errors) {
+      this.markRow(error.lineNumber - 1, 'line-red')
```
thomasjo: Change this class name from `line-red` to `latex-error`.

lib/error-marker.js
```diff
@@ -0,0 +1,33 @@
...
+    }
+    for (let warning of this.warnings) {
+      this.markRow(warning.lineNumber - 1, 'line-yellow')
```
thomasjo: Change this class name from `line-yellow` to `latex-warning`.

styles/markers.atom-text-editor.less
```diff
@@ -0,0 +1,8 @@
+:host(:host) .gutter .line-number {
+  &.line-red {
+    background-color: red;
```
thomasjo: We should not be picking colors directly. Use `@background-color-error` as a basis instead. My initial suggestion here would be to mimic the `highlight-*` classes as defined in the Atom styleguide. E.g.

```less
@import "ui-variables";
// ...
  &.latex-error {
    background-color: fade(@background-color-error, 50%);
    color: @text-color-error;
  }
// ...
```

styles/markers.atom-text-editor.less
```diff
@@ -0,0 +1,8 @@
...
+  }
+  &.line-yellow {
+    background-color: yellow;
```
thomasjo: Same as above, except use `@background-color-warning` and `@text-color-warning` instead.

Focus:
Done

thomasjo:
> Done

Excellent, but we also need to address the explicit color usage as I mentioned in https://github.com/thomasjo/atom-latex/pull/167#discussion_r54539563.

I can change it post-merge, but it would be great if you could refactor the style. Exactly what I want is this
```less
@import "ui-variables";

:host(:host) .gutter .line-number {
  &.latex-error {
    background-color: fade(@background-color-error, 50%);
    color: @text-color-error;
  }
  &.latex-warning {
    background-color: fade(@background-color-warning, 50%);
    color: @text-color-warning;
  }
}
```

Note that I haven't tested these styles, but they follow the official Atom styleguide, so I'm reasonably confident it will look OK.

Focus:
Sorry I completely missed that. I implemented the changes you asked for. I also found a small bug where editor.getPath() would return undefined when one of the editors is a new page and fixed this as well.

Here are some screenshots under various themes:
atom-dark:
atom-dark
atom-light:
atom-light
one-dark:
one-dark
seti:
seti

Focus:
I have given you access to my fork, so that you can push your own changes as well. I am more than happy to help with anything as well.

thomasjo:
This looks really good! I can make some minor tweaks after merging if necessary.
Thanks again for this, such an awesome feature :sparkles:

Focus:
Excellent. Thanks for feedback

PR #170, opened 2016-03-01
Add forward sync support to Atom PDF opener

nsaje:
I recently got a [PR](https://github.com/izuzak/atom-pdf-view/pull/111) merged into atom-pdf package which allows forward syncing. This patch provides the second part of the functionality that allows a user to forward sync to the Atom PDF viewer.

I don't write JS frequently so let me know if I should change anything :smiley:

### DIFF
--- a/lib/openers/atompdf-opener.js
+++ b/lib/openers/atompdf-opener.js
@@ -6,7 +6,7 @@ export default class AtomPdfOpener extends Opener {
   open (filePath, texPath, lineNumber, callback) {
     // Opens PDF in a new pane -- requires pdf-view module
 
-    function forwardSync(pdfView) {
+    function forwardSync (pdfView) {
       if (pdfView != null && pdfView.forwardSync != null) {
         pdfView.forwardSync(texPath, lineNumber)
       }
@@ -21,11 +21,8 @@ export default class AtomPdfOpener extends Opener {
       }
     }
 
-    const activePane = atom.workspace.getActivePane()
     // TODO: Make this configurable?
-    atom.workspace.open(filePath, {'split': 'right'}).done(function (pdfView) {
-      forwardSync(pdfView)
-    })
+    atom.workspace.open(filePath, {'split': 'right'}).done(forwardSync)
 
     // TODO: Check for actual success?
     if (callback) {
--- a/lib/openers/atompdf-opener.js
+++ b/lib/openers/atompdf-opener.js
@@ -5,18 +5,27 @@ import Opener from '../opener'
 export default class AtomPdfOpener extends Opener {
   open (filePath, texPath, lineNumber, callback) {
     // Opens PDF in a new pane -- requires pdf-view module
-    const openPanes = atom.workspace.getPaneItems()
-    for (const openPane of openPanes) {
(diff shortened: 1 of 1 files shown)


lib/openers/atompdf-opener.js
```diff
@@ -5,18 +5,27 @@ import Opener from '../opener'
 export default class AtomPdfOpener extends Opener {
   open (filePath, texPath, lineNumber, callback) {
     // Opens PDF in a new pane -- requires pdf-view module
-    const openPanes = atom.workspace.getPaneItems()
-    for (const openPane of openPanes) {
+
+    function forwardSync(pdfView) {
```
thomasjo: :art: There's a space missing before the opening paren.

lib/openers/atompdf-opener.js
```diff
@@ -5,18 +5,27 @@ import Opener from '../opener'
...
       // File is already open in another pane
-      if (openPane.filePath === filePath) { return }
+      if (openPaneItem.filePath === filePath) {
+        forwardSync(openPaneItem)
+        return
+      }
     }
 
-    const pane = atom.workspace.getActivePane()
+    const activePane = atom.workspace.getActivePane()
```
thomasjo: :art: This isn't used anymore after your changes, and can be removed.

lib/openers/atompdf-opener.js
```diff
@@ -5,18 +5,27 @@ import Opener from '../opener'
...
     // TODO: Make this configurable?
-    // FIXME: Migrate to Pane::splitRight.
-    const newPane = pane.split('horizontal', 'after')
-    // FIXME: Use public API instead.
-    atom.workspace.openURIInPane(filePath, newPane)
+    atom.workspace.open(filePath, {'split': 'right'}).done(function (pdfView) {
+      forwardSync(pdfView)
+    })
```
thomasjo: The inline function is unnecessary since you've defined the `forwardSync` function above. Replace with
```js
atom.workspace.open(filePath, {'split': 'right'}).done(forwardSync)
```

thomasjo:
Awesome news, and thanks for contributing! I left a few comments for things that need to be addressed :bow:

nsaje:
Here you go, let me know if there's anything else I need to address!

thomasjo:
Brilliant, thanks! You rock! :metal:

vikeri:
Seems like a great feature! How do I use it?

thomasjo:
> Seems like a great feature! How do I use it?

@vikeri Ensure your pdf-view package is up to date, then you should be good to go! Simply clicking on something in a pdf-view pane should trigger a sync in the .tex file, and invoking the *Sync* (or *Build*) command should trigger a sync in the PDF document.

If your results are not currently being opened in Atom, but in one of the platform-specific PDF applications (Preview or Skim on OS X, SumatraPDF on Windows), then you can force it use pdf-view via this package's settings. Look for *Always Open Result in Atom*.

vikeri:
Ok, I had some issues but now when I updated my Tex packages it works :) but there is no way to go from the `tex`-file to the corresponding place in the `pdf`-file?

PR #177, opened 2016-04-01
Migrate away from pure JavaScript Standard Style

thomasjo:
There are issues (e.g. arrow function parens) that can only be properly solved by using eslint plugins that cannot be used with pure Standard. Hence the move away from the `standard` package to `eslint-config-standard` together with `eslint-plugin-babel`.

Additional benefit is that globals can now be sensibly be scoped to folders again since we can use .eslintrc files and we can specify the `jasmine` env, etc :sparkles:

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -9,7 +9,7 @@
     "latex": true
   },
   "rules": {
-    "babel/array-bracket-spacing": 2,
+    "babel/array-bracket-spacing": [2, "always"],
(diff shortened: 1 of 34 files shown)


thomasjo:
There's little value in sticking to Standard JavaScript Style if we're going to deviate this much... investigating the benefits of going with something based on [Airbnb JavaScript Style Guide](https://github.com/airbnb/javascript) instead, coupled with [Flow](https://flowtype.org/).

PR #183, opened 2016-06-04
Add config option for SyncTeX

alexandrecormier:
I love this package but I'm not a huge fan of SyncTeX, so I added an option to let me disable it. I thought I'd submit a PR in case you're interested in merging it.

### DIFF
--- a/lib/builders/latexmk.js
+++ b/lib/builders/latexmk.js
@@ -33,13 +33,14 @@ export default class LatexmkBuilder extends Builder {
 
   constructArgs (filePath) {
     const outputFormat = atom.config.get('latex.outputFormat') || 'pdf'
+    const enableSynctex = atom.config.get('latex.enableSynctex') !== false
 
     const args = [
       '-interaction=nonstopmode',
       '-f',
       '-cd',
       `-${outputFormat}`,
-      '-synctex=1',
+      `-synctex=${enableSynctex ? 1 : 0}`,
       '-file-line-error'
     ]
 
--- a/lib/builders/latexmk.js
+++ b/lib/builders/latexmk.js
@@ -33,24 +33,26 @@ export default class LatexmkBuilder extends Builder {
 
   constructArgs (filePath) {
     const outputFormat = atom.config.get('latex.outputFormat') || 'pdf'
-    const enableSynctex = atom.config.get('latex.enableSynctex') !== false
 
     const args = [
       '-interaction=nonstopmode',
       '-f',
       '-cd',
       `-${outputFormat}`,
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -101,6 +101,12 @@
       "default": false,
       "order": 5
     },
+    "enableSynctex": {
+      "title": "Enable SyncTeX",
```
thomasjo: This should be `"description": `.

lib/builders/latexmk.js
```diff
@@ -33,13 +33,14 @@ export default class LatexmkBuilder extends Builder {
...
     const outputFormat = atom.config.get('latex.outputFormat') || 'pdf'
+    const enableSynctex = atom.config.get('latex.enableSynctex') !== false
 
     const args = [
       '-interaction=nonstopmode',
       '-f',
       '-cd',
       `-${outputFormat}`,
-      '-synctex=1',
+      `-synctex=${enableSynctex ? 1 : 0}`,
```
thomasjo: I think it would be nicer if this is enabled in a similar fashion to what we do for `enableShellEscape` below. I'll change how we apply the `outputFormat` option at some point to make everything consistent.

lib/builders/texify.js
```diff
@@ -31,10 +31,12 @@ export default class TexifyBuilder extends Builder {
   }
 
   constructArgs (filePath) {
+    const enableSynctex = atom.config.get('latex.enableSynctex') !== false
+
     const args = [
       '--batch',
       '--pdf',
-      '--tex-option="--synctex=1"',
+      `--tex-option="--synctex=${enableSynctex ? 1 : 0}"`,
```
thomasjo: See [the comment](https://github.com/thomasjo/atom-latex/pull/183/files#r65812979) for the equivalent change in latexmk.js.

thomasjo:
Thanks for contributing! 🙇 

Other than the few comment minor comments, this looks good. As soon as those comments are addressed, I'll merge 🚢 this.

package.json
alexandrecormier: Should it? I added a title to display proper casing. Like this:
2016-06-05-132017_136x33_scrot
Changing it to description yields this:
2016-06-05-131858_126x47_scrot

alexandrecormier:
Should be better now. 😄

package.json
thomasjo: Sorry, you are entirely correct! Was confusing myself thinking you were changing the description, but there is no description.

thomasjo:
Sorry about the delay in merging this! And thanks again for contributing 🙇

PR #184, opened 2016-06-12
Jobname support

yitzchak:
Add support for jobnames TeX Magic.  ``# !TEX jobnames = foo bar`` causes two builds. First with "foo" and second with "bar."

### DIFF
--- a/lib/composer.js
+++ b/lib/composer.js
@@ -41,7 +41,8 @@ export default class Composer {
     this.showProgressIndicator()
 
     return new Promise(async (resolve, reject) => {
-      let statusCode, result, jobnames = builder.getJobNamesFromMagic(rootFilePath)
+      let statusCode, result
+      let jobnames = builder.getJobNamesFromMagic(rootFilePath)
 
       const showBuildError = () => {
         this.showError(statusCode, result, builder)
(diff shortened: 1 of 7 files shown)


lib/composer.js
```diff
@@ -42,26 +42,29 @@ export default class Composer {
...
         reject(statusCode)
       }
 
       try {
-        statusCode = await builder.run(rootFilePath)
-        result = builder.parseLogFile(rootFilePath)
-        if (statusCode > 0 || !result || !result.outputFilePath) {
-          showBuildError(statusCode, result, builder)
-          return
+        for (let jobname of jobnames) {
```
thomasjo: This `let` can be made `const` since it is not mutated.

thomasjo:
This is looking really good!

Other than the little `const` comment I left, it would be awesome with a test that ensures the jobname loop works as expected. Essentially something that verifies that the builder is invoked the expected number of times. 

Having the aforementioned loop be an async loop would be a dream scenario, but I can sort that out at a later stage if you don't want to deal with that right now.

Thanks for helping out! 🙇

yitzchak:
:cool: Thanks for looking at it!

I fixed the let/const issue. I'll work on a test for the jobname loop. It might take me a bit longer to turn it into an async loop. I am not really a javascript expert.

thomasjo:
> It might take me a bit longer to turn it into an async loop. I am not really a javascript expert.

Don't worry about the async stuff. I can implement that at a later stage ⏩

thomasjo:
Sorry about the terribly long delay on merging this! I resolved the merge conflicts since it was my fault for letting this linger for such a long time.

Thanks again for contributing! I'll try to deal with your [other PR](https://github.com/thomasjo/atom-latex/pull/186) as soon as possible 🙇

yitzchak:
Awesome! Thanks for taking care of the conflicts and taking the PR.

Issue #91, opened 2015-06-01
Show a nice filtered view of the compilation log while compiling


LucFabresse:
I am just discovering atom and of course this LaTeX package.
Both are great and I want (will ;-)) to use it.
To me, the major obstacle today is the lack of feedback  during the compilation.
I've read that it has been discussed in other feature request but it deserves its own feature request.
 
A first step could be to just open a pane that contains the current compilation log when compiling.
A second step would be to filter the displayed log to only display important things with colors, .... 
A further step would be to allow the user to customize the filtering in preferences (e.g. not showing overflow, ...).
I also would like to have these kind of preferences:
- always displays compilation log (true/false)
- automatically close compilation log if no error (true/false)
...

Thanks for your work on this plugin.

WoodyWoodsta:
I could definitely have this too. Wading through the horrible latex log file is unpleasant! Let alone Latex's nonsensical error output as well :stuck_out_tongue_closed_eyes:

MaxNoe:
Do you know the `texfot` script that comes with texlive? No need to reinvent the wheel.

LucFabresse:
Thanks Max! I did not know that.
so the latex plugin should definitely use that ;-)

yitzchak:
Resolved by #232.

Issue #26, opened 2014-08-04
Implement log parsing


thomasjo:
- [x] Resolve output file.
- [x] Detect errors.
  - [x] Make sure we're detecting all possible errors.
- [x] Detect warnings.
- [ ] Display error and warning messages to the user in a friendly, convenient way.
  - [ ] Link error and warning messages to file and line number.

~~- [ ] If TeX file hasn't changed, parse log file instead?~~ (Requires way too much work due to root/master file support.)

anicetojunior:
+1

anicetojunior:
+1, this is a great feature...

thomasjo:
I've got some holiday coming up in a couple of weeks, so I'm hoping to tackle this issue, as well as #64, #93 and possible more.

thomasjo:
Thanks to #204, this has been greatly improved. Although there's still the task of making the whole experience more user-friendly.

PR #186, opened 2016-06-17
Display errors and warnings inline using Linter Indie interface

yitzchak:
Messages should show up after a build if linter is installed. I added a few more warning patterns to log-parser to help demonstrate the idea. This idea still needs a lot of work, but I thought it was worth a look now. If it seems interesting I can keep working on it. Also, if this is the primary notification method then I think that there could be a lot of code simplifications.


yitzchak:
Moved Linter stuff into linter-logger and added builder specific status code logging.

thomasjo:
Awesome stuff! I'm bogged down with Real Life™ work right now, but will find time over the next few days to review this, as well as #184.

yitzchak:
Completely understand. Let me know where I can fix or tweak anything! I have some time this summer.

yitzchak:
:see_no_evil:  The `constant.other.reference.latex` commit was incorrectly named. Should have been "resolve conflicts with master."

yitzchak:
After glancing through the open issues, I this PR might address #26 and #91.

yitzchak:
It looks like linter is about to have a major upgrade to v2 and change their message format. I have branches on my other linter-based project compatible that are with linter v2, so I am fairly certain of what would need to change in this PR to support v2. The legacy v1 message format in linter will still work, but it seems silly to rely on the legacy format since this is an un-merged PR.

I have been using this PR on both my home and work machines for over two months and have not run into any issues yet. Personally, I think it is very cool seeing LaTeX log messages show up in the source document.

I am just curious if you are still considering merging this PR? If so I can work on updating it to linter v2. I  understand either way.

lib/main.js
```diff
@@ -45,6 +45,14 @@ export default {
     this.composer.setStatusBar(statusBar)
   },
 
+  consumeLinter (indieRegistry) {
+    const myLinter = indieRegistry.register({name: 'LaTeX'})
+    this.disposables.add(myLinter)
+    // this.composer.setLinterIndie(myLinter)
```
thomasjo: Remove this commented out line.

spec/builders/latexmk-spec.js
```diff
@@ -3,6 +3,7 @@
 import helpers from '../spec-helpers'
 import path from 'path'
 import LatexmkBuilder from '../../lib/builders/latexmk'
+import * as _ from 'lodash'
```
thomasjo: This can be changed to `import _ from 'lodash'`.

spec/builders/latexmk-spec.js
```diff
@@ -92,6 +93,43 @@ describe('LatexmkBuilder', () => {
...
+          { type: 'Info', text: 'Class foo: Insignificant class issue' },
+          { type: 'Error', text: 'Package bar: Significant package issue' },
+          { type: 'Warning', text: 'Package bar: Package issue' },
+          { type: 'Warning', text: 'Package bar: Nebulous package issue' },
+          { type: 'Info', text: 'Package bar: Insignificant package issue' },
+          { type: 'Warning', text: 'There were undefined references' }
+        ]
+
+        for (const message of messages) {
+          expect(_.some(parsedLog.messages,
```
thomasjo: Should this be `_.all` instead of `_.some` to ensure all supported warnings and errors are correctly parsed?

thomasjo:
I'm happy to merge this, but there are a couple of things I'd like to see changed;

1. Combine the `showBuildWarning` and `showBuildWarning` config options into a `loggingLevel` option of type `enum`. It should have possible values `error`, `warn` and `info` with `warn` being the default.  
I'd ideally also like to see it moved to the end of schema.
2. I'm really worried about the complexity introduced into `log-parser.js`. With the compact regexes and varying number of capture groups etc, it's much more difficult to reason about that file now. While it might not be as pretty, I'd personally prefer multi-line patterns with EOL comments. It makes a little bit easier to spot capture groups etc.
3. My inline comments.
4. Specs for the linter stuff. I don't use Linter (and likely never will), so having automated tests that ensures the new functionality works is crucial. Especially considering all the refactoring and what not that thing undergoes.

Whether you want to update the Linter v2 stuff in this PR, or a new PR once this gets merged, is entirely up to you. Once you've addressed (to a reasonable extent) my concerns, I'm happy to merge this.

spec/builders/latexmk-spec.js
yitzchak: I used `_.some` because LaTeX font info messages will also show up and I am not sure if those  are entirely consistent across platforms. Since that that is not is not really what spec was trying to test I figured that looping through that messages that the file was known to generate and using `_.some` to find them was the easiest way to ignore the superfluous messages.

I have no problem thinking about it a little more.

thomasjo: OK, that makes sense! Maybe this could be changed to counting expected messages or some such? Not super important. If you leave it as is, it would be great if you could add a comment explaining the reasoning behind this structure.

yitzchak:
Sounds good. I'll start working on it for Linter v1 and see where they are once I finish addressing your concerns.

thomasjo:
Awesome stuff! Sorry about leaving this PR hanging for such a long time 🙇

yitzchak:
No worries mate! Your plugin is awesome and I heavily rely upon it.

BTW I am also on Slack if you ever need to discuss outside of the PRs.

package.json
```diff
@@ -22,7 +22,7 @@
   "devDependencies": {
     "babel-eslint": "^6.1.0",
     "snazzy": "^4.0.0",
-    "standard": "^7.1.2"
+    "standard": "^8.0.0"
```
thomasjo: Unless there's a good reason for bumping this in this PR, I'd rather leave this as is for now.

yitzchak: Understand. I was trying to fix the Appveyor build, but it seems to break Travis in the process. I'll roll it back.

yitzchak:
If linter is available then this PR automatically uses `LinterLogger` in preference over `ConsoleLogger`. Since `ConsoleLogger` is generally hidden and not all users may want to automatically use `LinterLogger` I am thinking that there should be a third option with selection of the logger up to the user.

yitzchak:
@thomasjo -> I am going to put this one back in the oven and let it bake :bread: for bit longer unless you have an objection. I do not think it is a bad idea to offer as logging option, but I think at a minimum it should be manually selected by the user.

thomasjo:
@yitzchak Sounds good to me. Let's ship #204, and then revisit this.

thomasjo:
Since we're letting this bake for a while, I'm closing it for the time being.

PR #191, opened 2016-07-09
Magic regex matches TeX and TEX

### LINKED ISSUE #139: Magic Comment Root File Search - Quotes and Upper/Lower Case of !TeX
```
# !TeX root = "main.tex"
```
does not work properly:

* !TeX instead of !TEX:
the magic comment will be ignored. Instead of the main file the current file will be build.
(most of latex magic comment tutorials i found are written with !TeX)
* Quotes around the filename:
  * build will build the main file and show the warning "LatTeX compilation error" without any error. (Click on the warning does not open anything)
  * clean does not seem to work. Nothing happens.

@redw0lf added the multi magic comment finder (thanks for that!)
Maybe he knows best how to implement that in his code?

I changed my magic comments based on that but it would be nice if that would work ;)

(tested with Ubuntu, latexmk)

yagop:
Closes  #139

### DIFF
--- a/lib/parsers/magic-parser.js
+++ b/lib/parsers/magic-parser.js
@@ -4,7 +4,7 @@ import fs from 'fs-plus'
 
 const magicCommentPattern = new RegExp('' +
   '^%\\s*' +    // Optional whitespace.
-  '!TEX' +      // Magic marker.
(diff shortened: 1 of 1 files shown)


thomasjo:
Thanks!

PR #192, opened 2016-07-18
Builder override

yitzchak:
Allow `%!TEX builder = ...` magic to override builder.


thomasjo:
It would be great if you could add a test (or two) for the new functionality. We need to specs to prevent regressions, etc.

yitzchak:
Got it. I'll make it happen! :bulb:

yitzchak:
I added a few tests, but I still have a few doubts about the functionality. I am planning on testing on some Windows machines next week.

spec/build-registry-spec.js
```diff
@@ -32,10 +32,22 @@ describe('BuilderRegistry', () => {
       expect(registry.getBuilder(filePath).name).toEqual('TexifyBuilder')
     })
 
-    it('throws an error when unable to resolve ambigious builder registration', () => {
+    it('throws an error when unable to resolve ambiguous builder registration', () => {
```
thomasjo: Nice catch! 😁

yitzchak:
Okay, I am satisfied. I tested in on MiKTeX with both `texify` and `latexmk` installed and was able to switch back and forth between the two with the magic comment.

I did have to move the saving of the file in `Composer` to before `getBuilder` was called to ensure that `MagicParser` would see a changed builder override originating from an unsaved file.

thomasjo:
LGTM! Feel free to hit that big, green button 👍

yitzchak:
Awesome! :boom:

PR #201, opened 2016-08-06
Add other grammar activation hooks

yitzchak:
[language-latex](https://atom.io/packages/language-latex) has not been updated for a least a year. The fork [language-tex](https://atom.io/packages/language-tex) is being actively updated and there is also an alternate package [language-latexsimple](https://atom.io/packages/language-latexsimple).

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,9 @@
     ]
   },
   "activationHooks": [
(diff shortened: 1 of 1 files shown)


thomasjo:
Thanks 🎉

PR #204, opened 2016-09-02
Improved Log Parsing and Error Reporting

yitzchak:
Remove `LinterLogger` from #186 thereby extracting the log parsing and builder specific error reporting logic. Make considering alternate loggers (linter-based, Atom notification, or other possibilities) a separate PR.


lib/logger.js
```diff
@@ -1,7 +1,57 @@
...
+
 export default class Logger {
-  error (/* statusCode, result, builder */) {}
-  warning (/* message */) {}
-  info (/* message */) {}
+  constructor () {
+    this.messages = []
+    this._group = null
+    const callback = () => { this.showFilteredMessages() }
+    atom.config.onDidChange('latex.loggingLevel', callback)
```
thomasjo: This `Disposable` should probably to be captured, and disposed of in a `destroy` method or some such. Right now this base class (and hence anything that extends it) is leaky.

Thinking about it, do we actually need to listen for this? We should only care about this when we parse the log file. If someone changes the logging level, that should only affect subsequent builds, not historical ones.

yitzchak: Good catch!

Initially, the reason that I was listening for it was in cause the user ran a build with the logging level set to "Error" then after the build completed, wanted to see the warnings or info messages already generated without re-running the build. That would make toggling the log level dynamically via a menu command possible.

I am okay with either. If we do not maintain the historical messages beneath the current logging level then I think that code that detects what the current logging level is, would probably need to move into the log parser.

yitzchak: The more that I think about it, I am leaning towards removing the dynamic filtering since it's meanless for `DefaultLogger`. Thoughts?

thomasjo: I'm always a fan of making things simpler, so this sounds good to me 👍

---
_Two of my favorite acronyms (and principles) are [KISS](https://en.wikipedia.org/wiki/KISS_principle), and [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)._

yitzchak: I agree! 894052f6816d99bea465bcc601299b974b694459 removes it without changing any logic.

lib/composer.js
```diff
@@ -34,34 +32,42 @@ export default class Composer {
...
     this.destroyErrorIndicator()
     this.showProgressIndicator()
 
     return new Promise(async (resolve, reject) => {
+      latex.log.group('LaTeX Build')
       const rootFilePath = this.resolveRootFilePath(filePath)
-      const jobnames = builder.getJobNamesFromMagic(rootFilePath)
-      for (const jobname of jobnames) {
-        try {
+      let jobnames = builder.getJobNamesFromMagic(rootFilePath)
```
thomasjo: This `let` can be `const`.

lib/error-marker.js
```diff
@@ -1,28 +1,19 @@
...
-    this.warnings = warnings
+    this.messages = messages
     this.markers = []
     this.mark()
   }
 
   mark () {
-    for (let error of this.errors) {
-      this.markRow(error.lineNumber - 1, 'latex-error')
+    for (let message of this.messages) {
```
thomasjo: `const`

lib/loggers/default-logger.js
```diff
@@ -0,0 +1,55 @@
...
+        case 'Warning':
+          console.warn(text)
+          break
+        default:
+          console.log(text)
+      }
+    }
+    console.groupEnd()
+
+    for (let editor of editors) {
```
thomasjo: `const`

package.json
```diff
@@ -109,6 +109,16 @@
       "default": true,
       "order": 6
     },
+    "loggingLevel": {
+      "type": "string",
+      "enum": [
+        "info",
+        "warning",
+        "error"
```
yitzchak: Fixed, and I added a description which is not so brief. I can remove the "such as ..." if you want to make it more terse.

thomasjo:
Thanks so much for this! 🎉

PR #206, opened 2016-09-03
Assign outputFilePath in resolveOutputFilePath

yitzchak:
Assignment of `outputFilePath` is missing in `resolveOutputFilePath`. I think this may have happened at the conversion from CoffeeScript to Javascript at 30a2dcb8a9a5e714ee026fc72ef741d7f578d46a


thomasjo:
Good catch! Thanks ❤️

PR #208, opened 2016-09-07
Add default environment for texify

### LINKED ISSUE #185: Package biblatex warning while using texify
I'm having troubles to build latex document using biblatex for my citations.
I use Windows, miktex2.9, texify as builder and pdflatex as engine.
Build does succeed without errors, but it throws warning:
Package biblatex Warning: Please (re)run Biber on the file: ...

Everything else in the final pdf is ok but citations is replaced with .bib file's keywords (Rivest1977, instead of Rivest et al. 1997)

I ran commands: `pdflatex doc && biber doc && pdflatex doc` directly from command line and pdf compiled as it should.

Is the problem with texify? When I'm using mac and latexmk as builder I have no problems.
Is it possible to fix this with atom-latex custom engine option? I'm not so sure how to use that
custom engine possibility.

yitzchak:
Only define `BIBTEX` variable if not defined system wide. Should fix #185

### DIFF
--- a/lib/builder.js
+++ b/lib/builder.js
@@ -45,8 +45,8 @@ export default class Builder {
     return new LogParser(logFilePath, texFilePath)
   }
 
-  constructChildProcessOptions (filePath) {
-    const env = _.clone(process.env)
+  constructChildProcessOptions (filePath, defaultEnv) {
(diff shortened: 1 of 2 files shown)


yitzchak:
Fixes #185

PR #214, opened 2016-09-11
Add DVI support in texify

yitzchak:
Fixes a few other small bugs discovered in the process of adding DVI support. I think we need to add CI support for `texify` at some point.


yitzchak:
The spec for this is now complete. Needs a review and we probably need to decide about `texify` support?

yitzchak:
Moot since we are deprecating `texify`.

PR #215, opened 2016-09-11
Create a logger that can help with issue reports

yitzchak:
This is still very rough around the edges. Mostly just the kernel of an idea of how to use `Logger` to record a filtered version of the LaTeX logs. We could certainly add other data to the report.

### DIFF
--- a/lib/latex.js
+++ b/lib/latex.js
@@ -27,6 +27,7 @@ export default class Latex {
     defineDefaultProperty(this, 'logger')
     defineDefaultProperty(this, 'opener')
 
+    this.observeLoggerConfig()
     this.observeOpenerConfig()
   }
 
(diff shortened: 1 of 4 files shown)


thomasjo:
My gut feeling is this functionality should not be a `Logger`. It doesn't really make sense for the user to configure their preferred logger in order to send us an issue.

I'm all in favor of capturing this information, but it needs a different approach. This should be implemented as something _"else"_, whatever that means. 99% of the time this will be caused by stuff that is not an issue with the package, but rather a typo on the LaTeX code, or other forms of end-user mistakes.

yitzchak:
I think you are right, it is going to get misused. I'll but it back in the oven! :bread:

yitzchak:
Closing since it will need to be somewhere. Will keep branch active to preserve code.

PR #218, opened 2016-09-11
Add MiKTeX and texify support in AppVeyor CI

yitzchak:
First get environment initialized and make sure all binaries can be found. `texify` support will be added once AppVeyor completes and succeeds with existing specs.

### DIFF
--- a/script/cibuild
+++ b/script/cibuild
@@ -17,6 +17,7 @@ import_dependencies() {
 
 exec_ci() {
   if [[ "${APPVEYOR:-}" == true ]]; then
+    Rscript -e "library(knitr)" -e "knit('spec\\fixtures\\knitr\\file.Rnw')"
     local script_path="/tmp/build-package.ps1"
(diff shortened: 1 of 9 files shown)


script/cibuild
```diff
@@ -27,8 +31,12 @@ exec_ci() {
 
 main() {
   import_dependencies
-  install_texlive
   install_knitr
+  if [[ "${TEX_DIST}" == "miktex" ]]; then
```
thomasjo: If we change this to `"${TEX_DIST:=texlive}"` we can remove the explicit defaults in `.travis.yml` and `.appveyor.yml`.
Unless you object, I'll tweak that while debugging the build.

yitzchak: Sounds great!

yitzchak:
It does not look the spec failure is due to MiKTeX package installation. I ran `mpm --admin --list` before and after the Rscript hack on a separate branch and did not see a difference in the list of installed packages.

thomasjo:
This should be good to go. We **have** to squash this when we merge — I do not want these experimental commits in the commit history 😝

Issue #220, opened 2016-09-12
Support for (u)pLaTeX


yudai-nkt:
I'm happy to see this package being actively updated recently and I think this is a good time to ask for a new feature.

As pdfTeX, XeTeX and LuaTeX do not have enough capability of Japanese typesetting without appropriate packages (e.g., [`zxjatype`](https://www.ctan.org/pkg/zxjatype) and [`luatex-ja`](https://www.ctan.org/pkg/luatexja)), the most prevailing method of Japanese LaTeX document is to use (u)pLaTeX and DVI driver (mainly dvipdfmx).

The current `latex` package can already interpret the magic comment like `% !TEX program = uplatex` to some extent, but also has some problems. Let me consider the following file `foo.tex`.

```latex
% !TEX program = uplatex
\documentclass[uplatex]{jsarticle}
\begin{document}
Hello world.
\end{document}
```

Compiling `foo.tex` on terminal by `uplatex foo.tex` goes successfully. On the other hand, invoking `uplatex` command on Atom by <kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>B</kbd> shows "LaTeX compilation error" dialog in the status bar although `foo.tex` does not contain any errors.

![foo](https://cloud.githubusercontent.com/assets/13291527/18428663/34b83388-790a-11e6-9a5b-2102906267c8.gif)

The other problem is that since (u)pLaTeX generate DVI files instead of PDF, we cannot make PDF with `latex` package when we use (u)pLaTeX. It would be nice to invoke DVI drivers such as `dvipdfmx` or `dvips` in combination with `platex`/`uplatex` within Atom.

I think that the number of (u)pTeX users is quite smaller than that of other modern TeX engines, but I will greatly appreciate it if these problems are addressed.

---

For reference, this is my minimal `.latexmkrc` for upLaTeX. Hope this helps.

```perl
$pdf_mode = 3; # generate pdf via dvipdfmx
$latex = "uplatex -kanji=utf8 -file-line-error -halt-on-error -interaction=nonstopmode %O %S";
$dvipdf = "dvipdfmx %O -o %D %S";
```

thomasjo:
@yudai-nkt Thanks for the fantastic, thorough feature request 🙇 

In summary you essentially want two features
1. Support for `platex` and `uplatex`.
2. The ability to pipe DVI output through e.g. `dvipdfmx`, `dvips`, etc.

[...]

PR #221, opened 2016-09-13
Add support for non-pdflatex engines

yitzchak:
This PR is an attempt to resolve some of the issues in #220 and also to fix some of the logic in `LatexmkBuilder` which makes it impossible to generate DVI files with engines such as `lualatex`. Still need to do more research on what other engines are not `pdflatex` derivatives and whether or not all `pdflatex` derivatives can generate DVI files.


yitzchak:
The following issues still need to resolved:

1. Write specs
2. Fix `outputFilePath` in `LogParser`. The DVI file gets opened because the LaTeX logs do not contain any information about the `dvipdf` run. That information is in the console output of `latexmk` or in the `.fdb_latexmk`.

In order to solve the second issue I think we need to parse either the console output or `.fdb_latexmk`. Maybe the builders should be responsible for choosing the appropriate log parser and return the log messages along with any generated files? Especially since `KnitrBuilder` does not have a LaTeX log. At a minimum let the builders return a list of generated files.

What do you think @thomasjo?

p.s. I'll do a squash to get rid of the senseless merge commit if you think this is a direction work pursuing.

yitzchak:
I added a FDB parser and allowed it to override the output path in `Builder.parseLogFile`. There may be a better or more logical way, but this avoids breaking the output path logic in `Composer`. The openers and SyncTex should now work.

lib/builders/latexmk.js
```diff
@@ -4,6 +4,8 @@ import childProcess from 'child_process'
 import path from 'path'
 import Builder from '../builder'
 
+const latexRegExp = /^latex|u?platex$/
```
thomasjo: I want to switch to uppercasing these module scoped constants. This should be renamed to `LATEX_PATTERN`.

lib/builders/latexmk.js
```diff
@@ -49,21 +51,22 @@ export default class LatexmkBuilder extends Builder {
...
     const args = [
       '-interaction=nonstopmode',
       '-f',
       '-cd',
-      `-${outputFormat}`,
       '-file-line-error'
     ]
 
+    const outputFormat = this.getOutputFormatFromMagic(filePath) ||
+      atom.config.get('latex.outputFormat') || 'pdf'
```
thomasjo: Extract this fall-through logic into a method. E.g. 
```js
getOutputFormat (filePath) {
  const outputFormat = this.getOutputFormatFromMagic(filePath)
  if (outputFormat) return outputFormat

  return atom.config.get('latex.outputFormat')
}
```

I don't understand why the `|| 'pdf'` part was added? We should use whatever default output format the user has configured? The default value is `pdf` by the way.

lib/builders/latexmk.js
```diff
@@ -49,21 +51,22 @@ export default class LatexmkBuilder extends Builder {
...
     const enableShellEscape = atom.config.get('latex.enableShellEscape')
     const enableSynctex = atom.config.get('latex.enableSynctex') !== false
-    const engineFromMagic = this.getLatexEngineFromMagic(filePath)
-    const customEngine = atom.config.get('latex.customEngine')
-    const engine = atom.config.get('latex.engine')
+    const engine = this.getLatexEngineFromMagic(filePath) ||
+      atom.config.get('latex.customEngine') ||
+      atom.config.get('latex.engine') || 'pdflatex'
+    const producer = this.getProducerFromMagic(filePath) ||
+      atom.config.get('latex.producer') || 'dvipdf'
```
thomasjo: Just as for the output format stuff, extract two methods here; e.g. `getLatexEngine` and `getProducer`.
Also try to avoid the inline truthy conditionals. It's generally much easier to read code that use explicit conditional blocks.

lib/builders/latexmk.js
```diff
@@ -75,12 +78,42 @@ export default class LatexmkBuilder extends Builder {
...
+            break
+        }
+      } else {
+        args.push(`-${outputFormat}`)
+      }
+    } else {
+      if (engine !== 'pdflatex') {
+        args.push(`-pdflatex="${engine}"`)
+      }
+      args.push(`-${outputFormat}`)
```
thomasjo: With these changes `constructArgs` is doing way too much. We need to extract some more granular logic. It's becoming really hard to read. Try to split into logical units... one example might be something like `getEngineArgs` which can also be split into smaller units (possibly one like `getPdfProducerArgs`).

lib/parsers/fdb-parser.js
```diff
@@ -0,0 +1,54 @@
...
+  getLines () {
+    if (!fs.existsSync(this.filePath)) {
+      throw new Error(`No such file: ${this.filePath}`)
+    }
+
+    const rawFile = fs.readFileSync(this.filePath, {encoding: 'utf-8'})
+    const lines = rawFile.replace(/(\r\n)|\r/g, '\n').split('\n')
+    return lines
+  }
+}
```
thomasjo: This is a perfect opportunity to introduce a `Parser` base class. We know have 3 different classes which share the same concerns; `LogParser`, `MagicParser` and `FdbParser`.

All share the same `getLines` method, and each should implement an "abstract" `parse` method.

package.json
```diff
@@ -161,70 +161,83 @@
       "default": "pdf",
       "order": 10
     },
+    "producer": {
+      "description": "Program to use when post-process DVI output in order to produce PDF. This is not used for pdflatex based builders. Currently only supported for latexmk.",
+      "type": "string",
+      "enum": [
+        "default",
```
thomasjo: What is default? Can we possibly be more opinionated and explicitly pick an explicit default producer instead?

package.json
```diff
@@ -161,70 +161,83 @@
       "default": "pdf",
       "order": 10
     },
+    "producer": {
+      "description": "Program to use when post-process DVI output in order to produce PDF. This is not used for pdflatex based builders. Currently only supported for latexmk.",
```
yudai-nkt: My comment is not related to the implementation, but I think using the word builder here is a bit confusing. It is already used to indicate latexmk and texify though the latter is going to be deprecated.

Also, XeTeX is not a pdfTeX based engine while LuaTeX is a hybrid successor of pdfTeX and [Omega](https://en.wikipedia.org/wiki/Omega_%28TeX%29). So something like "PDF-producible engines" is a more appropriate terminology IMHO (there might be even better words as I have poor English vocabulary).

lib/builders/latexmk.js
```diff
@@ -77,12 +80,35 @@ export default class LatexmkBuilder extends Builder {
...
+        switch (producer) {
+          case 'default':
+            // default is an option to protect a user's .latexmkrc from override
+            args.push('-pdfdvi')
+            break
+          case 'dvipdf':
+            args.push('-pdfdvi')
+            args.push('-e "\\$dvipdf = \'dvipdf %O %S %D\';"')
+            break
+          case 'dvipdfm':
```
yudai-nkt: As `dvipdfm` is a symlink to `dvipdfmx` since TeX Live 2009 and the driver file for `graphicx` package is removed in TL 2013, dvipdfm is regarded as obsolete for quite long. Details about compatibility between dvipdfm and dvipdfmx are a bit intricate, but in short, the use of dvipdfm with CJK documents is strongly discouraged unless under very limited circumstances. Hence, I think support for dvipdfm is an excessive burden that doesn't need to be imposed. I'm just letting you know because it's entirely up to you to decide whether or not to support such outdated distributions.

yitzchak: Thanks for the feedback @yudai-nkt!

It looks to me like both TeX Live and MiKTeX currently have `dvipdfmx` symlinked to `xdvipdfmx` so it is probably worth simplifying the list.

Cheers!

package.json
yitzchak: Default was meant to protect the user's `.latexmkrc` file in case they had a setting in there. I'll look at the latexmk code and see what order the evaluation happens in and whether this is really needed.

lib/builders/latexmk.js
thomasjo: Just wanted to add my thanks as well — great, insightful feedback, @yudai-nkt 🙇

package.json
yitzchak: Good point. We will see if we can straighten the language out.

lib/builders/latexmk.js
yudai-nkt: Just to make sure: I meant that `dvipdfm` is symlinked to `dvipdfmx`. In TeX Live 2016, `dvipdfmx` is symlinked to`xdvipdfmx`, as @yitzchak confirmed, and xdvipdfmx changes its behavior according to the extension of input file (`.dvi` or `.xdv`). That is not the case with TeX Live 2015 or earlier.

yitzchak: @yudai-nkt Got it. Makes sense.

lib/parsers/fdb-parser.js
yitzchak: The `getLines` is not the same in all three parsers. `FdbParser` and `LogParser` throw an error when the file is not present, whereas `MagicParser` just returns an empty array. Sub-classing from from the same `Parser` class causes spec failure no matter which default behavior is chosen.

I have a commit that will merge the parsers, but I had to wrap `getLines` in a try/catch in `MagicParser` in order to preserve the current behavior. We should probably make a decision about what behavior we prefer.

package.json
yitzchak: `default` :arrow_right: :fire:

lib/parsers/fdb-parser.js
yitzchak: Added a `defaultLines` argument to `getLines` to make this a little bit less of a hack.

lib/builders/latexmk.js
```diff
@@ -4,6 +4,8 @@ import childProcess from 'child_process'
 import path from 'path'
 import Builder from '../builder'
 
+const LATEX_PATTERN = /^latex|u?platex$/
```
yitzchak: Listing `latex` here is probably not needed. `latex` is currently symlinked to `pdflatex` as are most LaTeX engines that are listed as DVI producing engines. So far the only other one that I have been able to discover that cannot produce PDF is `lamed` which I believe is not really used anymore.

Issue #213, opened 2016-09-09
Builder override is broken by root file magic comment


yitzchak:
Because `getBuilder` is called before  the master LaTeX file path has been resolved the `builder` magic comment in the master file will not be used. Instead every LaTeX file would have to explicitly override the builder.

yitzchak:
Closing since we are deprecating the builder magic override.

Issue #211, opened 2016-09-09
Jobnames magic induced bugs


yitzchak:
SyncTeX forward doesn't know about jobnames and tries to open a pdf based on the source files name.

yitzchak:
`jobnames` magic has broken the clean command also.

yitzchak:
The output path caching is also broken by jobnames. For instance, when the magic is `%!TEX jobnames = foo` in the initial build and then changed to `%!TEX jobnames = bar` in successive builds, `foo.pdf` is still opened.

Issue #179, opened 2016-04-15
Add support for makeglossaries


mbellgb:
It would be great if there was a setting to allow the user to add extra commands during compilation, for instance the `makeglossaries` command.

thomasjo:
I fully agree. I've been contemplating the best way of implementing this for a little while now. Will try to find some time to document it either here, or in a placeholder PR.

thbtmntgn:
+1 !
Do you have some news @thomasjo ?

thomasjo:
@thbtmntgn Sadly this was put on The Backburner™, but now that I've implemented Knitr support, I'm feeling pretty confident about implementing some manner of pipeline system. Such a system might involve invoking `makeglossaries`, `makeindex`, Knitr, etc.

The biggest issue is deciding on how to let the user control/specify the pipeline. The simplest approach might be something akin to arara which simply uses comments. I want to begin by implementing the simplest possible solution though, so a full-blown pipeline concept might be too complex 🔮

sinh15:
Hi there,

Sorry to disturb you guys, I'm new to latex and quite super lost. 

I'm trying to build the glosseries through atom and although the functionality that you are discussing is still not in place, any idea on how could I approach the display of acronyms and glossaries in the meanwhile? Even if it's through console. Any input would be welcome!

Actually, running the command `\printglossaries{}` is displaying the Acronyms, but not the Glossary.

Sorry for disturbing in here.

yitzchak:
Support for `makeglossaries` will be in the next version. This is done by using `latexmk` custom rules. Also, support for Asymptote, MetaPost, `nomencl`, `index` and SageTeX is being added using the same mechanism.

Cheers!

PR #222, opened 2016-09-15
Migrate to centralized build state

yitzchak:
This PR is intended to start the process to migrate to a centralized build state. This is needed to make it possible to address #211 and #213, but also create the framework to explore solutions for #220 and #179.  A centralized build state might also help move the second stage builder resolution code out of `KnitrBuilder`.

Right now only the `BuildState` class has been added but not used anywhere as we want to move gradually in refactoring as we will need to rewrite specs, etc.

### DIFF
--- a/lib/build-state.js
+++ b/lib/build-state.js
@@ -46,6 +46,14 @@ export default class BuildState {
     this.activeJobname = value
   }
 
+  getActiveFilePath () {
+    return this.activeFilePath
+  }
+
+  setActiveFilePath (value) {
+    this.activeFilePath = value
+  }
+
   getLatexEngine () {
     return this.latexEngine
   }
--- a/lib/build-state.js
+++ b/lib/build-state.js
@@ -1,94 +1,55 @@
 /** @babel */
 
-import fs from 'fs-plus'
-import path from 'path'
-import MagicParser from './parsers/magic-parser'
-
-const masterFilePattern = new RegExp('' +
-  '^\\s*' +             // Optional whitespace.
-  '\\\\documentclass' + // Command.
-  '(\\[.*\\])?' +       // Optional command options.
-  '\\{.*\\}'            // Class name.
-)
-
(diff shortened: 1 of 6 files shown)


yitzchak:
I was looking at the `Latex` class and it gave me some ideas. It's not a completely "dumb" state class, but it removes all the magic parser and root file logic. What do you think @thomasjo?

```javascript
export default class BuildState {
  constructor (filePath) {
    this.filePath = filePath
    this.projectPath = path.dirname(filePath)
  }

  defineProperties (/* properties */) {
    for (const property of arguments) {
      const shadowProperty = `__${property}`
      const keyPath = `latex.${property}`

      Object.defineProperty(this, property, {
        get: function () {
          if (!this[shadowProperty]) {
            this[shadowProperty] = atom.config.get(keyPath)
          }
          return this[shadowProperty]
        },

        set: function (value) { this[shadowProperty] = value }
      })
    }
  }
}
```

One could call `defineProperties` from the constructor or after the class has been instantiated.

thomasjo:
We don't need to shadow any properties, etc. since we don't need any underlying "magic" default values. The build state should just be something simple like

```js
export default class BuildState {
  constructor (filePath, projectPath) {
    this.filePath = filePath
    this.projectPath = projectPath
  }

  getFilePath () {
    return this.filePath
  }

  getProjectPath () {
    return this.projectPath
  }

  getXyz () {
    return this.xyz
  }

  setXyz (value) {
    this.xyz = value
  }

  // ...
}
```

An alternative is to define properties with explicit getters and setters to control which ones are read-only etc. But I think just starting with method makes sense. If it becomes too verbose or cumbersome, we can always switch to properties.

yitzchak:
I am going to close this PR for the time being. I have made a start on this PR a few times enough to see what the obstacles are. I will probably start an experimental build pipeline PR eventually with the knowledge that I have gained.

Issue #162, opened 2016-01-15
SyncTeX for Evince


edwinksl:
I am trying to make SyncTeX work for the Evince PDF viewer. I have set the `Custom PDF Viewer Path` to `/usr/bin/evince`  and was hoping that SyncTeX would work out of the box but this isn't the case. Any tips for making SyncTex play well with Evince?

peterbabic:
Still good for you, I did set the same thing, but atom opens the resulting PDF in source textual form. I do not understand why. (using XFCE linux environment, Okular has too much KDE dependencies, so I do not want to use it).

yitzchak:
In order to enable SyncTeX with Evince we will probably need to have an opener class specifically for Evince. Currently, [SyncTeX support in Evince](https://help.gnome.org/users/evince/stable/index.html.en#synctex) works with gedit and can work with vim by using [evince_dbus.py](https://git.gnome.org/browse/gedit-plugins/tree/plugins/synctex/synctex/evince_dbus.py) which is supplied as a sample by gedit.

What do you think @thomasjo? Evince is the default viewer in Gnome 3 and hence my usual viewer, so I can certainly put it on my long term todo list, but supporting it will most likely result in a dbus dependency.

thomasjo:
I definitely see the benefit of adding this. My only concern is adding in a lot of additional complexity just to support a PDF viewer for a subset of users. If we can keep it fairly straight-forward, yet robust, then we should add support for Evince.

yitzchak:
Resolve #162

yitzchak:
This feature has been added and should be in the next release.

cbrnr:
How do I trigger forward and backward search? I can't seem to figure it out, neither Ctrl+click in Evince nor Ctrl+Alt+S in Atom is working...

yitzchak:
@cbrnr xdg-open is the fallback opener which will use Evince but not support SyncTeX. Trying putting the path to Evince in the settings. If that does not work, please open a new issue since this issue was to add support for Evince.

cbrnr:
@yitzchak according to the [Wiki](https://github.com/thomasjo/atom-latex/wiki/Supported-Viewers), Evince should support both forward and backward search. Evince is opening fine with the default settings, but I can't seem to figure out how fw/bw search works...

yitzchak:
[...]

PR #225, opened 2016-09-18
Add support for Evince with SyncTex

yitzchak:
This PR addresses #162. It is currently a bit of callback hell, but SyncTeX forward and backward are working.

## Remaining issues

- [x] Determine if `dbus.sessionBus()` needs to be disposed of. *It does not, although if we allow the user to change the opener we may need to call `removeListener`*
- [x] It should be possible to listen to all Evince instances and respond to `SyncSource` even if a build has not been done yet by listening to [add_signal_receiver](https://dbus.freedesktop.org/doc/dbus-python/api/dbus.bus.BusConnection-class.html#add_signal_receiver), but I have not been able figure out how do this using dbus-native. The current functionality matches that of okular, etc. *It's done with `addMatch` but it seems useless, since Evince does not start sending `SyncSource` until it receives `SyncView`.*
- [x] Improve the opener logic. Evince is only called if Okular is not present. Maybe the user should be able to choose? *I'll do it in another PR*
- [x] Test on old versions of Evince. dbus does not appear to be versioned like COM. The GNOME team added `timestamp` parameter to [`SyncView`](https://github.com/GNOME/evince/blob/d681da676c4a52590330426ba18196ddcfbb8037/shell/ev-gdbus.xml) in version [2.91.0](https://github.com/GNOME/evince/blob/master/NEWS). If we do not pass this parameter (zero works) then current versions break. Passing this parameter may cause old versions to break, so we may need to inspect the method signature. *Evince 2.91.0 was released in 2011, so I think we are okay assuming the `s(ii)u` signature.*
- [x] Investigate support for opening in foreground vs. background. *Done, although SyncView seems a little fickle about window activation so it may not be perfect.*

### DIFF
--- a/lib/openers/evince-opener.js
+++ b/lib/openers/evince-opener.js
@@ -4,78 +4,89 @@ import Opener from '../opener'
 import dbus from 'dbus-native'
 import url from 'url'
 
-function SyncSource (uri, point) {
+const EVINCE_OBJECT = '/org/gnome/evince/Evince'
+const EVINCE_APPLICATION_INTERFACE = 'org.gnome.evince.Application'
+
+const DAEMON_SERVICE = 'org.gnome.evince.Daemon'
+const DAEMON_OBJECT = '/org/gnome/evince/Daemon'
+const DAEMON_INTERFACE = 'org.gnome.evince.Daemon'
+
+const WINDOW_INTERFACE = 'org.gnome.evince.Window'
+
+function syncSource (uri, point) {
   atom.workspace.open(url.parse(uri).pathname).then(editor => editor.setCursorBufferPosition(point))
+}
+
+function getInterface (bus, serviceName, objectName, interfaceName) {
+  return new Promise((resolve, reject) => {
+    bus.getInterface(serviceName, objectName, interfaceName, (error, interfaceInstance) => {
+      if (error) {
+        reject(error)
+      } else {
+        resolve(interfaceInstance)
+      }
+    })
+  })
+}
+
+function getWindowList (application) {
+  return new Promise((resolve, reject) => {
+    application.GetWindowList((error, windowNames) => {
+      if (error) {
+        reject(error)
+      } else {
+        resolve(windowNames)
+      }
+    })
+  })
+}
+
+function findDocument (daemon, filePath) {
+  return new Promise((resolve, reject) => {
+    const uri = url.format({ protocol: 'file:', slashes: true, pathname: filePath })
+    console.log(uri)
+    daemon.FindDocument(uri, true, (error, documentName) => {
+      console.log('sdf')
+      if (error) {
+        reject(error)
+      } else {
+        resolve(documentName)
+      }
+    })
+  })
 }
 
 export default class EvinceOpener extends Opener {
   constructor () {
     super()
     this.bus = dbus.sessionBus()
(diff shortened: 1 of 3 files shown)


lib/openers/evince-opener.js
```diff
@@ -0,0 +1,81 @@
+/** @babel */
+
+import Opener from '../opener'
+import dbus from 'dbus-native'
+import url from 'url'
+
+function SyncSource (uri, point) {
```
thomasjo: Unless I'm missing something, this should be `syncSource`

lib/openers/evince-opener.js
```diff
@@ -0,0 +1,81 @@
...
+  open (filePath, texPath, lineNumber, callback) {
+    this.getWindow(filePath, texPath, (error, windowInterface) => {
+      if (!error) {
+        // The window interface does not seem to work with file protocol
+        windowInterface.SyncView(texPath, [lineNumber, 0], 0)
+      }
+      if (callback) callback(error)
+    })
+  }
+}
```
thomasjo: We really need to hide all the callback stuff behind promises. Let's move to the async future ⚡

lib/openers/evince-opener.js
```diff
@@ -0,0 +1,81 @@
...
+  constructor () {
+    super()
+    this.bus = dbus.sessionBus()
+    this.daemon = null
+    this.windows = {}
+  }
+
+  getDaemon (callback) {
+    if (this.daemon) return callback(null, this.daemon)
+    this.bus.getInterface('org.gnome.evince.Daemon', '/org/gnome/evince/Daemon', 'org.gnome.evince.Daemon',
```
thomasjo: I don't understand what these strings mean. Can we extract them as constants with meaningful names?

lib/openers/evince-opener.js
```diff
@@ -0,0 +1,81 @@
...
+
+    const _getWindow = (name, windowName) => {
+      this.bus.getInterface(name, windowName, 'org.gnome.evince.Window', (error, windowInterface) => {
+        if (error) return callback(error)
+        this.windows[texPath] = windowInterface
+        windowInterface.on('SyncSource', SyncSource)
+        windowInterface.on('Closed', () => delete this.windows[texPath])
+        // Hack: a small wait seems to help when the document is opened for the
+        // first time.
+        setTimeout(() => callback(null, windowInterface), 200)
```
thomasjo: I really don't like this. Is there some way we can avoid this? Using some kind of polling maybe? This is almost guaranteed to cause problems under heavy workloads, on a slow computer, or what not.

lib/openers/evince-opener.js
yitzchak: You are not. I'll fix it.

lib/openers/evince-opener.js
yitzchak: They are the object and interface names. I attach constants to them.

lib/openers/evince-opener.js
yitzchak: I agree. I am going to remove it and see what happens.

lib/openers/evince-opener.js
yitzchak: Fixed in latest commit.

PR #227, opened 2016-09-20
Deprecate texify support

thomasjo:
This PR officially deprecates our `texify` support. The user is presented with a fairly annoying warning notification. Clicking the button opens the wiki page ["Using-latexmk-with-MiKTeX"](https://github.com/thomasjo/atom-latex/wiki/Using-latexmk-with-MiKTeX).

file1_tex_ __users_thomasjo_scratch_latex_test1

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,16 +12,32 @@ from the command line.
 
 ## Prerequisites
 ### TeX distribution
-Since this package relies upon `latexmk`, a reasonably up to date and working
-TeX distribution is required. The only current officially supported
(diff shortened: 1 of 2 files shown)


yitzchak:
How long do you think the deprecation period should be?  If it is more then a couple minor versions maybe we should add a note about texify being feature frozen.

thomasjo:
> How long do you think the deprecation period should be?

I'm thinking mid-October.

PR #231, opened 2016-09-27
MiKTeX Portable CI

yitzchak:
This PR is an attempt to use a custom MiKTeX portable distribution to speed up the AppVeyor CI for MiKTeX.


yitzchak:
@thomasjo There appears to be a bug either in the choco package of MiKTeX or the particular options passed to the MiKTeX setup utility. Christian Schenk was unable to reproduce so I went back to the drawing board with this PR.... and success!!!

PR #238, opened 2016-10-02
Add rebuild command

### LINKED ISSUE #112: Improve build process
Right now we're solely relying on `latexmk` (and similarly `texify` in the case of MiKTeX) to determine when to do a rebuild when invoking the *Build* command. This leads to problems if a user changes the package's engine configuration, since this does not trigger a rebuild, and thus makes it seem as if changing e.g. the engine from `pdflatex` to `lualatex` has no effect (see #110).

As I see it, there are two viable solutions to this, and both requires observing changes to package configuration. If the configuration has changed in such a way as to warrant a rebuild, we can
 1. automatically invoke the *Clean* command before proceeding with the execution of the *Build* command, or
 2. introduce a new *Rebuild* command, and when a user invokes *Build* we can redirect the execution to the new rebuild workflow.

I think option (2) is the best approach, as this would not be dependent upon configuration of the *Clean* command to be "correct" and hence be more rigorous. In addition I think it makes sense to have a *Rebuild* command.

yitzchak:
Rebuild command (`-g` option) is only available from menu or forced by any changes to the `latex` config. 

- [x] Investigate using sets to track which documents need a rebuild

Resolves #112.

### DIFF
--- a/lib/composer.js
+++ b/lib/composer.js
@@ -75,7 +75,6 @@ export default class Composer {
     const jobs = jobnames.map(jobname => this.buildJob(builder, rootFilePath, jobname, shouldRebuild))
 
     return Promise.all(jobs)
-      .catch(error => latex.log.error(error.message))
       .then(() => {
         this.destroyProgressIndicator()
         latex.log.groupEnd()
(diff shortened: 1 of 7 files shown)


lib/composer.js
```diff
@@ -60,41 +63,43 @@ export default class Composer {
...
+    const jobnames = builder.getJobNamesFromMagic(rootFilePath)
+
+    if (this.rebuildCompleted && !this.rebuildCompleted.has(rootFilePath)) {
+      shouldRebuild = true
+      this.rebuildCompleted.add(rootFilePath)
+    }
+
+    const jobs = jobnames.map(jobname => this.buildJob(builder, rootFilePath, jobname, shouldRebuild))
+
+    return Promise.all(jobs)
```
thomasjo: Does this make sense given that you're handling errors and successes in the containing function? I'm not 100% sure, but the only errors that might be raised on the calling side will originate from inside the `then` block. E.g. if either `this.destroyProgressIndicator()` or `latex.log.groupEnd()` results in an exception being raised.

Let's return `true/false`, or move the `then/catch` code to the call site?

yitzchak: I think that makes sense.

yitzchak: Can you check out the current commit? I still need to call `latex.log.groupEnd()` to push the messages out, but the `try/catch` should guarantee no promise rejection.

PR #241, opened 2016-10-05
Add clean option to latexmk

### LINKED ISSUE #58: Include missfont.log in clean
When using the Clean command, the file `missfont.log` isn't removed.

I think that file should be removed too..

Thanks

### LINKED ISSUE #161: Clean does not clean subdirectories (using latexmk)
When including tex-files of subdirectories with

    \include{subdir/Chapter1}

and using the latexmk engine, at least all aux-files still remain in this subdirectory after cleaning.

To reproduce this, use my small test repo [orzechow/latexmk-clear-test](https://github.com/orzechow/latexmk-clear-test).

At first I thought it might be a latexmk bug and actually encoutered one with exactly the same behaviour. This latexmk bug has been confirmed, but as I saw in the code, atom-latex does not even call latexmk for cleaning. (Why not actually?)
So it seems to be an issue of atom-latex.

### LINKED ISSUE #211: Jobnames magic induced bugs

yitzchak:
This is an experimental PR to see what the effect of adding an `action` parameter to `Builder.run` would be. Currently action is: `build`, `rebuild` or `clean`. I may add `full-clean`.

This resolves #58 and resolves #161, plus partial for #211 for which I believe it is possible to completely fix without resorting to a build state solution.

### Issues
- [x] Do we even need a `cleanExtensions` setting? `latexmk` seems very intelligent about cleaning. I have changed the default setting to `[]`. *I would say yes*
- [x] Do we need a `latex:full-clean` command? Our current command removed SyncTeX and output files which is really a full clean operation. *Added*
- [x] Specs
  - [x] Clean on simple LaTeX file
  - [x] Clean on simple LaTeX file with jobname
  - [x] Clean on simple LaTeX file with output directory
  - [x] Clean on complex LaTeX file
  - [x] Full clean on simple LaTeX file
  - [x] Full clean on simple LaTeX file with jobname
  - [x] Full clean on simple LaTeX file with output directory
  - [x] Full clean on complex LaTeX file
- [x] Document `latexmk`
  - [x] Options
  - [x] Logic
  - [x] Internal configuration
- [x] Test scenarios
  - [x] `minted` - `$clean_ext` supports globbing, but adds `%R` if not included. `latexmk` is never going to clean something like `_minted` folders because they are created outside of `pdflatex` through `runsystem`, etc.
  - [x] [orzechow/latexmk-clear-test](https://github.com/orzechow/latexmk-clear-test) - `latex:clean` removes all but SyncTeX and output. `latex:full-clean` removes everything.
- [x] Fixes #211

### DIFF
--- a/lib/builders/latexmk.js
+++ b/lib/builders/latexmk.js
@@ -2,6 +2,7 @@
 
 import path from 'path'
 import Builder from '../builder'
+import { BUILD_ACTION, REBUILD_ACTION, CLEAN_ACTION } from '../actions'
 
 const LATEX_PATTERN = /^latex|u?platex$/
 
@@ -12,7 +13,7 @@ export default class LatexmkBuilder extends Builder {
     return path.extname(filePath) === '.tex'
   }
 
-  async run (filePath, action, jobname) {
+  async run (filePath, action = BUILD_ACTION, jobname = null) {
     const args = this.constructArgs(filePath, action, jobname)
     const command = `${this.executable} ${args.join(' ')}`
     const options = this.constructChildProcessOptions(filePath, { max_print_line: 1000 })
@@ -43,7 +44,7 @@ export default class LatexmkBuilder extends Builder {
     }
   }
 
-  constructArgs (filePath, action, jobname) {
+  constructArgs (filePath, action = BUILD_ACTION, jobname = null) {
     const args = [
       '-interaction=nonstopmode',
       '-f',
@@ -58,10 +59,10 @@ export default class LatexmkBuilder extends Builder {
     const cleanExtensions = atom.config.get('latex.cleanExtensions')
 
     switch (action) {
-      case 'rebuild':
+      case REBUILD_ACTION:
         args.push('-g')
         break
-      case 'clean':
+      case CLEAN_ACTION:
         args.push('-c')
         if (cleanExtensions.length) {
           args.push(`-e "\\$clean_ext='${cleanExtensions.join(' ')}'"`)
--- a/lib/builders/latexmk.js
+++ b/lib/builders/latexmk.js
@@ -2,7 +2,7 @@
 
 import path from 'path'
 import Builder from '../builder'
-import { BUILD_ACTION, REBUILD_ACTION, CLEAN_ACTION } from '../actions'
+import { BUILD_ACTION, REBUILD_ACTION, CLEAN_ACTION, FULL_CLEAN_ACTION } from '../actions'
 
 const LATEX_PATTERN = /^latex|u?platex$/
 
@@ -56,17 +56,16 @@ export default class LatexmkBuilder extends Builder {
     const enableShellEscape = atom.config.get('latex.enableShellEscape')
     const enableSynctex = atom.config.get('latex.enableSynctex') !== false
     const engine = this.getLatexEngine(filePath)
-    const cleanExtensions = atom.config.get('latex.cleanExtensions')
 
     switch (action) {
       case REBUILD_ACTION:
         args.push('-g')
         break
       case CLEAN_ACTION:
-        args.push('-c')
-        if (cleanExtensions.length) {
-          args.push(`-e "\\$clean_ext='${cleanExtensions.join(' ')}'"`)
-        }
+        args.push(this.constructCleanArgs('-c'))
+        break
+      case FULL_CLEAN_ACTION:
+        args.push(this.constructCleanArgs('-c'))
         break
     }
 
@@ -97,6 +96,13 @@ export default class LatexmkBuilder extends Builder {
 
     args.push(`"${filePath}"`)
     return args
+  }
+
+  constructCleanArgs (option) {
+    const cleanExtensions = atom.config.get('latex.cleanExtensions')
+    if (!cleanExtensions.length) return option
+
+    return `${option} -e "\\$clean_ext='${cleanExtensions.join(' ')}'"`
   }
 
   constructPdfProducerArgs (filePath) {
--- a/lib/builders/latexmk.js
+++ b/lib/builders/latexmk.js
@@ -12,8 +12,8 @@ export default class LatexmkBuilder extends Builder {
     return path.extname(filePath) === '.tex'
   }
 
-  async run (filePath, jobname, shouldRebuild) {
-    const args = this.constructArgs(filePath, jobname, shouldRebuild)
+  async run (filePath, action, jobname) {
+    const args = this.constructArgs(filePath, action, jobname)
     const command = `${this.executable} ${args.join(' ')}`
     const options = this.constructChildProcessOptions(filePath, { max_print_line: 1000 })
 
@@ -43,7 +43,7 @@ export default class LatexmkBuilder extends Builder {
     }
   }
 
-  constructArgs (filePath, jobname, shouldRebuild) {
+  constructArgs (filePath, action, jobname) {
     const args = [
       '-interaction=nonstopmode',
       '-f',
@@ -55,10 +55,20 @@ export default class LatexmkBuilder extends Builder {
     const enableShellEscape = atom.config.get('latex.enableShellEscape')
     const enableSynctex = atom.config.get('latex.enableSynctex') !== false
     const engine = this.getLatexEngine(filePath)
+    const cleanExtensions = atom.config.get('latex.cleanExtensions')
 
-    if (shouldRebuild) {
-      args.push('-g')
+    switch (action) {
+      case 'rebuild':
+        args.push('-g')
+        break
+      case 'clean':
+        args.push('-c')
+        if (cleanExtensions.length) {
+          args.push(`-e "\\$clean_ext='${cleanExtensions.join(' ')}'"`)
+        }
+        break
     }
+
     if (jobname) {
       args.push(`-jobname=${jobname}`)
     }
@@ -95,9 +105,9 @@ export default class LatexmkBuilder extends Builder {
       case 'ps2pdf':
         return '-pdfps'
       case 'dvipdf':
-        return '-pdfdvi -e "\\$dvipdf = \'dvipdf %O %S %D\';"'
+        return '-pdfdvi -e "\\$dvipdf=\'dvipdf %O %S %D\';"'
       default:
-        return `-pdfdvi -e "\\$dvipdf = \'${producer} %O -o %D %S\';"`
+        return `-pdfdvi -e "\\$dvipdf=\'${producer} %O -o %D %S\';"`
     }
   }
 }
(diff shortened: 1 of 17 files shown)


thomasjo:
I can't recall exactly why I moved away from the clean option, but there was a good reason for it. 
One immediate downside is that it is no longer possible to simply perform a clean, regardless of building.

If we rely solely on the latexmk's clean going forward (and remove the `latex:clean` command), we will have to verify that we cover _all_ scenarios currently supported by having the command. Does it correctly clean up everything that the current system does? This should be covered by specs to guard against regressions.

lib/builders/latexmk.js
```diff
@@ -50,7 +50,7 @@ export default class LatexmkBuilder extends Builder {
     }
   }
 
-  constructArgs (filePath, jobname, shouldRebuild) {
+  constructArgs (filePath, action, jobname) {
```
thomasjo: What about changing this to
```js
constructArgs (filePath, action = 'build', jobname = null)
```

Then we can remove the `action` value from the specs. Even if you leave those, I think it makes sense to provide defaults here. What do you think?

lib/composer.js
```diff
@@ -12,28 +12,24 @@ export default class Composer {
     this.configChange = atom.config.onDidChange('latex', () => {
       this.rebuildCompleted = new Set()
     })
+    this.outputLookup = new Map()
```
thomasjo: Not super important, but this can be moved outside of the constructor (e.g. above);
```js
outputLookup = new Map()

constructor () { // ...
```

If we move it, we should move the `builderRegistry` property as well 🤘

lib/builders/latexmk.js
yitzchak: I think that sounds good.

lib/builders/latexmk.js
```diff
@@ -62,10 +62,20 @@ export default class LatexmkBuilder extends Builder {
...
+    switch (action) {
+      case 'rebuild':
+        args.push('-g')
+        break
+      case 'clean':
+        args.push('-c')
+        if (cleanExtensions.length) {
+          args.push(`-e "\\$clean_ext='${cleanExtensions.join(' ')}'"`)
+        }
+        break
```
yitzchak: A deep clean and then a rebuild can be done with the `-gg` option. We could add a command for that.

thomasjo: Yeah, I realize that, but I'm just questioning whether it makes sense. Also `-gg` is not strictly equivalent to what we're supporting now. What we have now is a "normal" clean (can be configured to be equivalent to `-gg` by the user) which can then be followed by a build. If the user wants to control when the do a `-c` style clean (our current default) they can. With the new scheme, they don't have that control anymore.

I'm not saying having this functionality is vital or even makes sense, I'm simply arguing the fact we are changing the current behavior. We reduce code complexity somewhat (if we ignore the magic `action` strings), but we also lose some control and flexibility.

yitzchak:
I agree completely about writing complete specs and really testing this functionality out. I am also planning on documenting the inner logic of `latexmk`. As a start, below is the command line options available.

| Option | Command         | Behavior                                   |
|:-------|:----------------|:-------------------------------------------|
| None   | `latex:build`   | Regular build                              |
| `-g`   | `latex:rebuild` | Build ignoring timestamps                  |
| `-gg`  | None            | Deep clean, then build ignoring timestamps |
| `-c`   | `latex:clean`   | Clean and do **not** build                 |
| `-C`   | None            | Deep clean and do **not** build            |

thomasjo:
Ahaa, I never really read the docs well enough to realize the clean flags to **not** run builds. Guess I have should have read more carefully;

> [...] This and the other clean-ups are instead of a regular make.

yitzchak:
### `latexmk` clean configuration variables

| Name                                 | Default | Description                                                                    |
|:-------------------------------------|:-------:|:-------------------------------------------------------------------------------|
| `$clean_ext`                         | `""`    | extensions of files to be deleted beyond standard set                          |
| `$clean_full_ext`                    | `""`    | extensions of files to be deleted when doing cleanup_full, beyond `$clean_ext` |
| `$cleanup_fdb`                       | `0`     | Removal of file for latexmk's file-database                                    |
| `$cleanup_includes_generated`        | `0`     | Deletes files generated by (pdf)latex (found from \openout lines in log file). |
| `$cleanup_includes_cusdep_generated` | `0`     | Deletes files generated by custom dependencies                                 |

yitzchak:
### `latexmk` Clean Logic

1. Get a list of generated files. First try the FDB, then the log and lastly
just use extensions. Group the list into `$index_bibtex_generated`,
`$aux_files`, and `$other_generated`.
2. If BibTeX mode is off remove `bbl` from the generated extention list.
3. Delete any files found in the auxilary directory that have an extension of
`fdb_latexmk`, `blg`, `ilg`, `log`, `aux.bak`, `idx.bak` or any extension that
appears in `$clean_ext` or `@generated_exts`. Globbing enabled but `%R` is
prepended if it is not in the pattern. In other words, the file must contain
the root file name or job name as applicable.
4. Remove `texput.log`, `texput.aux`, `missfont.log` and any files in
`$index_bibtex_generated` or `$aux_files`.
5. Remove the dependents file if it was enabled.
6. Remove everything in `$other_generated` if `$cleanup_includes_generated` is
enabled.
7. If `$cleanup_includes_cusdep_generated` is enabled then remove all files
generated by custom dependencies.
8. If full clean is enabled then remove files from the output directory that
have an extension of `dvi`, `dviF`, `ps`, `psF`, `pdf`, `synctex.gz` or any
extension in `$clean_full_ext`. Same globbing rules as step 3.

yitzchak:
### File Extension Comparison

| Extension     | `latexmk`                               | `atom-latex` |
|:-------------:|:----------------------------------------|:-------------|
| `aux`         | Cleaned in step 3.                      | Cleaned      |
| `bcf`         | Cleaned in step 3.                      | Not cleaned  |
| `bbl`         | Cleaned in step 3 if BibTeX is enabled. | Cleaned      |
| `blg`         | Cleaned in step 3.                      | Cleaned      |
| `fdb_latexmk` | Cleaned in step 3.                      | Cleaned      |
| `fls`         | Cleaned in step 3.                      | Cleaned      |
| `idx`         | Cleaned in step 3/4.                    | Not cleaned  |
| `ind`         | Cleaned in step 3/4.                    | Not cleaned  |
| `lof`         | Cleaned in step 3.                      | Cleaned      |
| `log`         | Cleaned in step 3.                      | Cleaned      |
| `lol`         | Not cleaned unless in `$clean_ext`      | Cleaned      |
| `lot`         | Cleaned in step 3.                      | Cleaned      |
| `out`         | Cleaned in step 3.                      | Cleaned      |
| `nav`         | Not cleaned unless in `$clean_ext`      | Cleaned      |
| `pdf`         | Cleaned in step 8.                      | Cleaned      |
| `snm`         | Not cleaned unless in `$clean_ext`      | Cleaned      |
| `synctex.gz`  | Cleaned in step 8.                      | Cleaned      |
| `toc`         | Cleaned in step 3.                      | Cleaned      |

yitzchak:
@thomasjo With the last commit this PR will clean at least as well as our current algorithm and probably better in most cases without user configuration. I have not set `$cleanup_includes_generated` or `$cleanup_includes_cusdep_generated`. Setting those would probably be appropriate in a full clean.

The specs are decent, but not exhaustive since testing things like BibLaTex or Beamer will probably require modifying the TeX Live/MiKTeX CI builds.

I have documented the inner logic and configuration of the clean function of `latexmk` and it seems to be very safe since it will never delete files that it cannot prove were generated by LaTeX and do not have the job name in the path (with a few hard coded exceptions). The logic seems slightly convoluted but reasonable. Then again, we might end up with very similar logic if we parse the FDB and use it to try to make intelligent decisions about cleaning.

In summary, I think we probably have enough information to make a decision once we have time to talk.

yitzchak:
Closing since we have decided to go with an improved version of our own algorithm.

harmsk:
Why did you decide to abandon this approach? This action approach seems like a really good idea.

This approached works really nicely with people who must collaborate with others who do not use Atom. We can still get the same behavior because they could just call `latexmk -c` at the command line. Having the same behavior between collaborators seems like a good idea even if we don't all use the same editors/tools.

Further, this approach gave me the chance to make atom-latex work with legacy projects that use makefiles. (https://github.com/harmsk/atom-latex/tree/make-builder) For these projects, there is little incentive to change how we build documents. Being able to call `make all` and `make clean` is really helpful from within Atom.

Please let me know your thoughts. I'm really interesting in understanding your thoughts behind this and #253.

yitzchak:
@harmsk Closing this PR was not so much about abandoning an "action" approach to building as it was about our approach to the cleaning algorithm.

@thomasjo may jump in with his own thoughts and perspectives as I cannot speak to those directly.

For myself, I came to the conclusion that the cleaning algorithm of `latexmk` is not as flexible or configurable as I think it should be. There are open issues with `atom-latex` that I think a cleaning algorithm should be able to address, and it does not appear that `latexmk -c` has the capabilities to resolve those issues.

We have some longer term goals and ideas that may result in revisiting an "action" approach. For example, unified build state or build "pipelines." In other words, the idea is not dead yet.

thomasjo:
@harmsk The reason we decided to improve our current cleaning logic instead of going with this PR, is that it gives us much greater flexibility and control. Basically by using our own approach we can do novel things such as path globbing and placeholder substitution in the cleaning patterns. This gives us both a great default pattern set by which we clean, while also giving the user a lot of freedom in configuring their own set of patterns to suit their specific needs.

Also, if we decide to revisit support for alternate builders in the future (e.g. `arara`), we simply cannot rely on `latexmk`, which would mean introducing a custom/alternate cleaning scheme either way.

At the end of the day, this package is, and always will be, opinionated in some sense. It is not intended to fulfill everyone's desires and wishes, but it should work well for the vast majority of users. And perhaps most importantly, it should work well for most users out of the box. Personally I feel that we probably have too much configurability already (read: not enough opinions), but that's a discussion for another time and place.

**TL;DR** The approach implemented in this PR is not flexible enough for what we wanted/needed to achieve.

thomasjo:
Just as an additional follow-up; at some point after we've reached v1.0 (which is not far off) we might consider exposing some services that could allow third-party packages to provide custom functionality for concepts such as cleaning, opening PDF/DVI, rendering log messages, etc.

The core functionality needs to be more stable before we go down that route however. And this is not a promise it will happen, just a hint at what might come in the not too distant future, and which would allow us or the community to provide alternate functionality that is tangential to the default experience.

harmsk:
Thanks for the thorough responses! I agree. This makes complete sense. `latexmk -c` does miss quite a bit.

When/Where is the discussion/planning happening for alternate builders? I would like to try to get this makefile builder working relatively soon, if possible. Too many LaTeX projects use them!

Issue #236, opened 2016-09-30
Terminate stuck pdflatex instances


Nauhaie:
Sometimes pdflatex gets stuck in a loop (due to mistakes in some .tex files), and a new instance is launched at every new attempt to compile, without terminating the previous one. We end up with dozens of pdflatex instances running at the same time and eating up all the CPU...

yitzchak:
Thanks for feature request! We'll look into it.

yitzchak:
Feature has been added and will be in next release.

PR #244, opened 2016-10-06
Add Process Manager

yitzchak:
:bulb: This PR is the kernel of an idea for a method of resolving #236.

- [x] Add kill command to menu
- [x] Kill at beginning of build
- [x] Add tooltip to status bar status
- [x] Add kill command to status bar
- [x] Specs

### DIFF
--- a/lib/process-manager.js
+++ b/lib/process-manager.js
@@ -1,0 +1,30 @@
+/** @babel */
+
+import childProcess from 'child_process'
+import kill from 'tree-kill'
+
+export default class ProcessManager {
+  processes = new Set()
+
+  exec (command, options) {
+    return new Promise((resolve) => {
+      const proc = childProcess.exec(command, options, (error, stdout, stderr) => {
+        this.processes.delete(proc.pid)
+        resolve({
+          statusCode: error ? error.code : 0,
+          stdout,
+          stderr
+        })
+      })
+      this.processes.add(proc.pid)
+    })
+  }
+
+  kill () {
+    for (const pid of this.processes.values()) {
+      kill(pid)
+    }
+    this.processes.clear()
+  }
+}
+
--- a/lib/process-manager.js
+++ b/lib/process-manager.js
@@ -1,7 +1,11 @@
 /** @babel */
 
+import _ from 'lodash'
 import childProcess from 'child_process'
 import kill from 'tree-kill'
+import { promisify } from './werkzeug'
+
+const killTree = promisify(kill)
 
 export default class ProcessManager {
   processes = new Set()
@@ -20,10 +24,8 @@ export default class ProcessManager {
     })
   }
 
-  kill () {
-    for (const pid of this.processes.values()) {
-      kill(pid)
-    }
+  async kill () {
+    await Promise.all(_.map(this.processes.values(), pid => killTree(pid)))
     this.processes.clear()
   }
 }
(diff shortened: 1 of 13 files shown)


lib/process-manager.js
```diff
@@ -0,0 +1,29 @@
...
+          statusCode: error ? error.code : 0,
+          stdout,
+          stderr
+        })
+      })
+      this.processes.add(pid)
+    })
+  }
+
+  kill () {
```
thomasjo: Maybe rename this to `killRunningProcesses` or something along those lines.

lib/process-manager.js
```diff
@@ -0,0 +1,29 @@
+/** @babel */
+
+import childProcess from 'child_process'
+import kill from 'tree-kill'
+
+export default class ProcessManager {
+  processes = new Set()
+
+  exec (command, options) {
```
thomasjo: Rename this to something like `spawnChildProcess`, `executeChildProcess`, or `startChildProcess`?

lib/views/status-indicator.js
```diff
@@ -43,4 +46,14 @@ export default class StatusIndicator {
     this.properties = properties
     return etch.update(this)
   }
+
+  readAfterUpdate () {
```
thomasjo: I don't immediately understand how this name refers to what is happening inside of it. Rename it? Let's embrace verbose function names 🤘

spec/process-manager-spec.js
```diff
@@ -0,0 +1,53 @@
...
+import fs from 'fs-plus'
+import temp from 'temp'
+import path from 'path'
+import ProcessManager from '../lib/process-manager'
+
+describe('ProcessManager', () => {
+  let processManager, tempPath
+
+  function constructCommand (fileName) {
+    const filePath = path.join(tempPath, fileName)
```
thomasjo: Let's pass in `tempPath` as an argument to the function, and move the function outside of the `describe` block?

It doesn't seem likely to be a problem anytime soon, but if tests were suddenly executed in parallel, we would potentially be affected by a race condition here by relying on the captured `tempPath` to not be changed by another thread/process.

spec/process-manager-spec.js
```diff
@@ -0,0 +1,53 @@
...
+
+      waitsFor(() => killed, 5000)
+    })
+
+    it('kills old latexmk instances, but not ones created after the kill command', () => {
+      let oldKilled = false
+      let newKilled = false
+
+      processManager.exec(constructCommand('old.tex')).then(result => {
+        oldKilled = true
```
thomasjo: I think it would look nicer if you put this block (and the equivalent below) on a single line;

```js
processManager.exec(constructCommand('old.tex')).then(() => { oldKilled = true })
```

lib/process-manager.js
yitzchak: How about `killChildProcesses` to be consistent with `executeChildProcess`?

lib/views/status-indicator.js
yitzchak: It's the required name for reading the DOM after etch update.

thomasjo: Aha, thanks for the clarification 🙇

lib/process-manager.js
thomasjo: Sounds good!

PR #253, opened 2016-10-11
Improve clean logic

### LINKED ISSUE #58: Include missfont.log in clean

### LINKED ISSUE #161: Clean does not clean subdirectories (using latexmk)

yitzchak:
This PR is attempt to improve the clean logic, make it more flexible and yet still remaining safe. The current logic is:

1. Get a list of files from the FDB and any files from the output directory that start with the job name. This becomes the `generatedFiles`.
2. Loop through the list of patterns in `cleanPatterns`.
3. Do named property replacement on each pattern. Currently, `{jobname}`, `{output}` or `{name}`.
4. If the pattern is an absolute path (`/minted-{jobname}`) use it as a glob on the file system after joining to the root path.
5. If the pattern is a relative path, (`**/*.aux`) use it as a glob pattern on `generatedFiles` after joining to the root path.
6. :zap: any matches from step 4 or step 5.

This resolves #58 and resolves #161


lib/composer.js
```diff
@@ -170,6 +170,7 @@ export default class Composer {
 
     for (const pattern of patterns) {
       const absolutePattern = path.join(rootPath, pattern)
+      console.log(`${pattern} -> ${absolutePattern}`)
```
thomasjo: 🔥

lib/composer.js
```diff
@@ -5,7 +5,9 @@ import { shell } from 'electron'
 import fs from 'fs-plus'
 import path from 'path'
 import BuilderRegistry from './builder-registry'
-import { getEditorDetails, heredoc } from './werkzeug'
+import { getEditorDetails, heredoc, replaceProperitiesInString } from './werkzeug'
```
thomasjo: Typo here; should be `replacePropertiesInString`.

lib/composer.js
```diff
@@ -154,33 +153,80 @@ export default class Composer {
...
+    for (const file of files.values()) {
+      fs.removeSync(file)
     }
+  }
 
-    if (jobname) {
-      name = jobname
+  getCleanPatterns (builder, rootFilePath, jobname) {
+    const { name, ext } = path.parse(rootFilePath)
+    const output = atom.config.get('latex.outputDirectory')
```
thomasjo: I really think we should change `output` to `outputDir`. This applies to all other instances of `output`.

lib/werkzeug.js
```diff
@@ -30,5 +30,9 @@ export default {
     const lineNumber = position.row + 1
 
     return { editor, filePath, position, lineNumber }
+  },
+
+  replaceProperitiesInString (text, properties) {
```
thomasjo: Same typo here obviously; should be `replacePropertiesInString`.

lib/composer.js
yitzchak: Nice catch.

PR #254, opened 2016-10-12
Support synctex for Knitr files

### LINKED ISSUE #251: synctex does not work with knitr
If an Rnw file is build with the concordance option and the synctex.gz file is patched, then ctrl-alt-s should work for Rnw files.

Attached is a working example: [knitr_synctex.tar.gz](https://github.com/thomasjo/atom-latex/files/522508/knitr_synctex.tar.gz)

Please note that because atom-latex does not yet support the concordance option and patching, you will need to do this with the Makefile in the example.

To test this out:
```sh
# install R library patchSynctex
make rSnapshot

# compile pdf
make
```

Now try to use synctex in Atom; ctrl-alt-s in the Rnw file. It should take you to the location in pdf-view. It doesn't do anything.

If you add Rnw as a file extension to this function in composer.js
```javascript
  isTexFile (filePath) {
    // TODO: Improve will suffice for the time being.
    return !filePath || filePath.search(/\.(tex|lhs|Rnw)$/) > 0
  }
```
It works. But I'm not sure this is the best way to fix this.

harmsk:
Check for knitr files in sync command. resolves #251

### DIFF
--- a/lib/composer.js
+++ b/lib/composer.js
@@ -123,7 +123,7 @@ export default class Composer {
 
   async sync () {
     const { filePath, lineNumber } = getEditorDetails()
-    if (!filePath || !(this.isTexFile(filePath) || this.isKnitrFile(filePath))) {
+    if (!filePath || !this.isTexFile(filePath)) {
       return
     }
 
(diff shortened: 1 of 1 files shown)


lib/composer.js
```diff
@@ -250,6 +250,11 @@ export default class Composer {
     return !filePath || filePath.search(/\.(tex|lhs)$/) > 0
```
thomasjo: Let's simply change this regex to
```js
/\.(tex|lhs|[rs]nw)$/i
```

lib/composer.js
```diff
@@ -250,6 +250,11 @@ export default class Composer {
...
   }
 
+  isKnitrFile (filePath) {
+    // scope: text.tex.latex.knitr
+    return !filePath || filePath.search(/\.(Rnw|Snw|rnw|snw)$/) > 0
+  }
+
```
thomasjo: Remove this function. I realize you were told to add this, but I don't see why it's necessary.

yitzchak:
@thomasjo I am on it. I'll make it happen.

yitzchak:
Thanks @harmsk!

PR #256, opened 2016-10-12
Add concordance and patchSynctex to KnitrBuilder

yitzchak:
if `patchSynctex` is not available a warning is logged. Also, setting the output directory option appears to break `patchSynctex` so a warning for that might in order.


yitzchak:
@harmsk Concordance and patchSynctex have been merged. Thanks for the suggestion!

PR #262, opened 2016-10-15
add builder for makefiles

harmsk:
A number of older, legacy, and complex projects use Makefiles and this would be really helpful for working with these documents in Atom.

Would you consider adding something like this to build LaTeX documents with make?

### DIFF
--- a/lib/builders/make.js
+++ b/lib/builders/make.js
@@ -1,0 +1,48 @@
+/** @babel */
+
+import path from 'path'
+import fs from 'fs'
+import Builder from '../builder'
+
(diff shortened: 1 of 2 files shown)


thomasjo:
First of all, thanks so much for taking the time to contribute! We really appreciate the effort!

Unfortunately we have no desire to provide support for "legacy make projects" in the OOTB experience. As mentioned in a comment not long ago, I really want to add some extensibility points once we hit v1.0 (which is not far away) for exactly this type of functionality.

Sadly I cannot offer any workaround other than maintaining a fork until those bits ship.

Once again, thanks for your time and effort 🙇

PR #224, opened 2016-09-18
Improve test coverage

thomasjo:
This PR is a direct extension of #216.

---
- [x] `KnitrBuilder` has OK coverage, but **should** be improved.
  - [x] `canProcess`.
- [x] `LatexmkBuilder` has OK coverage, but **should** be improved.
  - [x] `canProcess`.
  - [x] `logStatusCode`.
  - [x] All permutations supported by `constructArgs`.
- [x] `DefaultLogger` has zero coverage, which **must** be improved.
  - [x] `showErrorMarkersInEditor`
  - [x] `sync`
- [x] `FdbParser` has reasonable coverage, but **can** be improved.
- [ ] `LogParser` has reasonable coverage, but **can** be improved.
- [x] `BuilderRegistry` has poor coverage, which **must** be improved.
  - [x] `getAllBuilders`.
- [ ] `Composer` has less than ideal coverage, which **should** be improved.
  - [x] `sync` — might need too much mocking to be worth it however.
  - [x] `moveResult`.
  - [ ] `resolveOutputFilePath`.
  - [ ] `showResult`
- [ ] `Latex` has OK coverage, but the tests **should** be improved.
  - [ ] Status bar indicator functionality.

---

Resolves #216

PR #265, opened 2016-10-16
Add Opener Registry

### LINKED ISSUE #85: Consider missing opener executable as an error?
See https://github.com/thomasjo/atom-latex/pull/83#issuecomment-104665130.

I think the suggestion makes sense, to consider a missing opener executable (or ill-configured opener path) as an error. Maybe this could/should also be tied into #64?

### LINKED ISSUE #235: Some viewers are not DVI/PS compatible
When output format is set to DVI we try to open in the default viewer even though it may not be able to open DVI files. For instance `pdf-view` tries to open no matter what the file extension is.

| Viewer                                                              | PDF  | PS   | DVI  | Linux | MacOS | Win32 |
|:--------------------------------------------------------------------|:----:|:----:|:----:|:-----:|:-----:|:-----:|
| [Evince](https://wiki.gnome.org/Apps/Evince)                        | :ok: | :ok: | :ok: |  :ok: |       |  :ok: |
| [Okular](https://okular.kde.org/)                                   | :ok: | :ok: | :ok: |  :ok: |  :ok: |  :ok: |
| [pdf-view](https://atom.io/packages/pdf-view)                       | :ok: |      |      |  :ok: |  :ok: |  :ok: |
| [Preview](https://support.apple.com/en-us/HT201740)                 | :ok: | :ok: |      |       |  :ok: |       |
| [Skim](http://skim-app.sourceforge.net/)                            | :ok: | :ok: | :ok: |       |  :ok: |       |
| [Sumatra PDF](http://www.sumatrapdfreader.org/free-pdf-reader.html) | :ok: |      |      |       |       |  :ok: |

### LINKED ISSUE #252: Error when PDF viewer is already open
If I build my latex file with my PDF viewer (SumatraPDF) already open then I get the following error "An error occurred while trying to run opener (1)." When the error is displayed it doesn't show errors or warnings from the latex log file. The PDF file is created and updated in Sumatra and the window does become selected. If I turn off the option "Open result after successful build" or close Sumatra before building then I no longer get the error and can see the warnings from the Latex log file.

yitzchak:
This is PR implements a simple opener registry which tries to pick an opener based on availability and features. The user can override the automatic selection via the `latex.opener` setting.

### Algorithm
1. If a specific opener has been set with `latex.opener` and that one can open the file then use it.
2. Filter out the openers that cannot open the requested file.
3. Abort and show a warning message if there no openers left after the filter.
4. Sort the openers in decreasing priority with SyncTeX capability ranked higher then "open in background" capability.
5. If `latex.enableSynctex` is enabled then use the viewer with the highest priority that supports SyncTeX if one is present. If there are openers that support both SyncTeX and "open in background" then these will be at the front of the list because of the priority sort.
6. If `latex.openResultInBackground` is enabled then use the viewer with the highest priority that supports opening in background if one is present.
7. If all else fails, just use the highest priority opener.

### Issues
- [x] Migrate `latex.alwaysOpenResultInAtom` setting to `latex.opener = 'pdf-view'` setting.
- [x] Simplify settings? Checking the existance of openers could be done with calls like `okular --version` if we replaced the `latex.texPath` with `latex.paths`. Would make all the opener path settings unneeded and make it possible to find 64bit SumatraPDF without user configuration. *Maybe later*
- [x] Refactor `EvinceOpener` to move methods into class.
- [x] Specs

Resolves #235 and resolves #85 and resolves #252.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -8,10 +8,10 @@ export default {
     this.disposables = new CompositeDisposable()
 
     this.disposables.add(atom.commands.add('atom-workspace', {
-      'latex:build': () => this.composer.build(false),
-      'latex:rebuild': () => this.composer.build(true),
-      'latex:clean': () => this.composer.clean(),
-      'latex:sync': () => this.composer.sync(),
+      'latex:build': () => composer.build(false),
+      'latex:rebuild': () => composer.build(true),
+      'latex:clean': () => composer.clean(),
+      'latex:sync': () => composer.sync(),
       'latex:kill': () => latex.process.killChildProcesses(),
       'latex:sync-log': () => latex.log.sync(),
       'core:close': () => latex.log.hide(),
@@ -23,7 +23,7 @@ export default {
         // Let's play it safe; only trigger builds for the active editor.
         const activeEditor = atom.workspace.getActiveTextEditor()
         if (editor === activeEditor && atom.config.get('latex.buildOnSave')) {
-          this.composer.build()
+          composer.build()
         }
       }))
     }))
@@ -35,9 +35,9 @@ export default {
       delete this.disposables
     }
 
-    if (this.composer) {
-      this.composer.destroy()
-      delete this.composer
+    if (global.composer) {
+      composer.destroy()
+      delete global.composer
     }
 
     if (global.latex) {
@@ -54,13 +54,12 @@ export default {
   },
 
   bootstrap () {
-    if (this.composer && global.latex) { return }
+    if (global.composer && global.latex) { return }
 
     const Latex = require('./latex')
     const Composer = require('./composer')
 
     global.latex = new Latex()
-    latex.opener.initializeOpeners()
-    this.composer = new Composer()
+    global.composer = new Composer()
   }
 }
--- a/lib/main.js
+++ b/lib/main.js
@@ -60,6 +60,7 @@ export default {
     const Composer = require('./composer')
 
     global.latex = new Latex()
+    latex.opener.initializeOpeners()
     this.composer = new Composer()
   }
 }
--- a/lib/main.js
+++ b/lib/main.js
@@ -49,7 +49,7 @@ export default {
     this.bootstrap()
     latex.status.attachStatusBar(statusBar)
     return new Disposable(() => {
-      latex.status.detachStatusBar()
+      if (latex) latex.status.detachStatusBar()
     })
   },
 
--- a/lib/opener-registry.js
+++ b/lib/opener-registry.js
@@ -21,6 +21,10 @@ export default class OpenerRegistry {
   }
 
   async open (filePath, texPath, lineNumber) {
+    const name = atom.config.get('latex.opener')
+    const openResultInBackground = atom.config.get('latex.openResultInBackground')
+    const enableSynctex = atom.config.get('latex.enableSynctex')
+
     let openers = _.filter(this.openers, opener => opener.canOpen(filePath))
 
     if (!openers.length) {
@@ -30,9 +34,19 @@ export default class OpenerRegistry {
 
     let opener
 
-    const name = atom.config.get('latex.opener')
     if (name !== 'automatic') {
       opener = _.find(openers, opener => opener.getName() === name)
+    }
+
+    if (!opener && enableSynctex) {
+      // If the user wants openResultInBackground also and there is an opener
+      // that supports that and SyncTeX it will be the first one because of
+      // the priority sort.
+      opener = _.find(openers, opener => opener.hasSynctex())
+    }
+
+    if (!opener && openResultInBackground) {
+      opener = _.find(openers, opener => opener.canOpenInBackground())
     }
 
     if (!opener) opener = openers[0]
--- a/lib/opener-registry.js
+++ b/lib/opener-registry.js
@@ -1,0 +1,53 @@
+/** @babel */
+
+import _ from 'lodash'
+import fs from 'fs-plus'
+import path from 'path'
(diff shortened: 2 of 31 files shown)


lib/opener-registry.js
```diff
@@ -0,0 +1,56 @@
...
+import _ from 'lodash'
+import fs from 'fs-plus'
+import path from 'path'
+
+export default class OpenerRegistry {
+  constructor () {
+    this.createOpeners()
+  }
+
+  async createOpeners () {
```
thomasjo: Maybe rename this to `initializeOpeners` or some such.

lib/opener-registry.js
```diff
@@ -0,0 +1,56 @@
...
+    const moduleDir = path.join(__dirname, 'openers')
+    const entries = fs.readdirSync(moduleDir)
+    this.openers = entries.map(entry => {
+      const OpenerImpl = require(path.join(moduleDir, entry))
+      return new OpenerImpl()
+    })
+    // Sort the openers into decreasing priority
+    this.openers = _.orderBy(this.openers, [opener => opener.hasSynctex(), opener => opener.canOpenInBackground()], ['desc', 'desc'])
```
thomasjo: Extract this "capability list" into a separate variable for readability? Call it `capabilityPrioritization` or something along those lines. Maybe simply `capabilities`?

lib/opener-registry.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  async open (filePath, texPath, lineNumber) {
+    const name = atom.config.get('latex.opener')
+    const openResultInBackground = atom.config.get('latex.openResultInBackground')
+    const enableSynctex = atom.config.get('latex.enableSynctex')
+
+    let openers = _.filter(this.openers, opener => opener.canOpen(filePath))
```
thomasjo: `const openers = this.openers.filter(opener => opener.canOpen(filePath))`

Also, I'd probably name it `candidates` instead of `openers` to indicate this is the list of candidates from which we will select an opener.

lib/opener-registry.js
```diff
@@ -0,0 +1,56 @@
...
+
+    if (!openers.length) {
+      latex.log.warning(`No opener found that can open ${filePath}.`)
+      return
+    }
+
+    let opener
+
+    if (name !== 'automatic') {
+      opener = _.find(openers, opener => opener.getName() === name)
```
thomasjo: `opener = openers.find(opener => opener.getName() === name)`

yitzchak:
One solution to the `latex.opener` settings and `getName` function issue would be to rename the openers to be consistent with the names in `latex.opener`. This means `atompdf-opener.js` :arrow_right: `pdf-view.js` and `xdg-opener.js` :arrow_right: `xdg-open.js`. We then load using:

```javascript
const schema = atom.config.getSchema('latex.opener')
const dir = path.join(__dirname, 'openers')
const ext = '.js'
this.openers = new Map()
for (const name of schema.enum) {
  const OpenerImpl = require(path.format({ dir, name, ext ))
  this.openers.set(name, return new OpenerImpl())
}
```

The selection logic would then be:

1. If a specific opener has been set with `latex.opener` and that one can open the file then use it.
2. Sort the openers in decreasing priority with SyncTeX capability ranked higher then "open in background" capability.
3. Filter out the openers that cannot open the requested file.
4. Abort and show a warning message if there no openers left after the filter.
5. If `latex.enableSynctex` is enabled then use the viewer with the highest priority that supports SyncTeX if one is present. If there are openers that support both SyncTeX and "open in background" then these will be at the front of the list because of the priority sort.
6. If `latex.openResultInBackground` is enabled then use the viewer with the highest priority that supports opening in background if one is present.
7. If all else fails, just use the highest priority opener.

thomasjo:
I like that idea!

yitzchak:
@thomasjo Can you give it one last review? I have added specs for `OpenerRegistry` and avoided needing to do package activation by initializing the openers only in package activation. Maybe not the best solution, but it works.

lib/opener-registry.js
```diff
@@ -0,0 +1,63 @@
+/** @babel */
+
+import _ from 'lodash'
+import path from 'path'
+
+export default class OpenerRegistry {
+  openers = new Map()
+
+  async initializeOpeners () {
```
thomasjo: Why is this async?

lib/main.js
```diff
@@ -60,6 +60,7 @@ export default {
     const Composer = require('./composer')
 
     global.latex = new Latex()
+    latex.opener.initializeOpeners()
```
thomasjo: Can we move this to the `Latex` constructor if/when we make `initializeOpeners` synchronous? Feels like some responsibility has leaked out here — this needs to be done either in `Latex` or `OpenerRegistry`.

yitzchak: It was done in `OpenerRegistry` but that caused problems in the spec since the schema was not available.

thomasjo: Then I propose we fix the spec instead. This just doesn't feel right 💭

yitzchak: Ok. I'll have to dig into the package activation issue to make it happen. `OpenerRegistry` is getting created in a whole bunch of places in the specs. Moving it back won't cause spec failure, just console messages, BTW.

yitzchak:
@thomasjo I've tried migrated the specs to use package activation by making the most minimal changes to specs unrelated to this PR. Looks like everything is working aside from latexmk 4.48 being broken on MiKTeX. I think we should do further tweaks to the unrelated specs if needed in #224 to avoid too many merge conflicts. In other words, I think is basically ready to :ship:!

Issue #243, opened 2016-10-06
Not building: "latexmk: Bad command line arguments"


yourfriendgerasim:
Just tried to use atom-latex for the first time. Building documents fails with the error message: `latexmk: Bad command line arguments` in the "LaTeX Log" on the bottom of the Atom screen. Package settings are left default. 

I'm on Windows 7, MiKTeX 2.9, Atom 1.10.2, atom-latex 0.37.0, latexmk 4.22d.

`latexmk` run from a command prompt works just normal. 

Any ideas?

yitzchak:
Can you please verify that the latexmk version is indeed 4.22d? If so you probably need to upgrade since 4.22d was released in 2011. There was a whole bunch of options added in 4.31 include LaTeX passthrough options.

[Latexmk: current and previous versions](http://personal.psu.edu/jcc8/software/latexmk-jcc/versions.html)

yourfriendgerasim:
Yes, it is indeed that version. 

    >latexmk -v
    Latexmk, John Collins, 26 January 2011. Version 4.22d

I will ask my admin for an upgrade and report if that fixes the issue.

yitzchak:
@yourfriendgerasim Any luck?

yitzchak:
Closing due to known issues with old versions of `latexmk`. Will all check for version.

yourfriendgerasim:
Upgrading fixed the issue.

yitzchak:
@yourfriendgerasim Awesome. I'm glad it's fixed.

RanSaSK:
So how did you finally solve your problem? I have the same problem. I'm curious.

Issue #54, opened 2015-02-09
Detect if `latexmk` can be found


thomasjo:
In other words, detect if `texPath` setting is correctly configured and/or a TeX distibution has been installed. I guess this is a two-pronged feature;

- [ ] Detect if `latexmk` can be found.
- [ ] Attempt to detect if a TeX distribution has been installed,
  - [ ] and auto-configure the `texPath` setting?

nscaife:
@thomasjo I can help with the Windows support if needed.

nschneid:
FYI, I am using MacTeX-2008, which does not include `latexmk`. I was able to install it manually from the download at https://www.ctan.org/pkg/latexmk/?lang=en. Perhaps suitable instructions can be given for this case.

PR #276, opened 2016-10-23
Check environment at startup

### LINKED ISSUE #62: Add "getting started" tests on first launch
As an enhancement, it might be worthwhile to add a getting started page on the first launch of the package.  I'm thinking of a basic checklist-driven workflow that checks:
- `latexmk` found
- PDF reader found
- basic compilation successful (perhaps include a short tex file in the package?)

yitzchak:
Expands the `checkEnvironment` function to do other checks besides deprecation notices and version migration. Since information checks are done also the messages are placed in the log versus using Atom notifications.

Current checks:
1. `latexmk` version check. Helps warn users about #243 situations. Also verifies that `latexmk` can be run outside of a build. [Error severity]
1. `Rscript` versions check. Mostly just to make sure `Rscript` can be run. [warning severity]
2. `knitr` and `patchSynctex` library check. [warning severity]
3. PDF capable viewer check. [error severity]
4. PS capable viewer check. [warning severity]
5. DVI capable viewer check. [warning severity]

Addresses some of #54 (does not automatically configure TeX path).

Resolves #62.

### DIFF
--- a/lib/builder-registry.js
+++ b/lib/builder-registry.js
@@ -1,6 +1,5 @@
 /** @babel */
 
-import _ from 'lodash'
 import fs from 'fs-plus'
 import path from 'path'
 import MagicParser from './parsers/magic-parser'
@@ -22,11 +21,11 @@ export default class BuilderRegistry {
     return (BuilderImpl != null) ? new BuilderImpl() : null
   }
 
-  async checkEnvironment () {
+  async checkRuntimeDependencies () {
     const builders = this.getAllBuilders()
     for (const BuilderImpl of builders) {
       const builder = new BuilderImpl()
-      await builder.checkEnvironment()
+      await builder.checkRuntimeDependencies()
     }
   }
 
@@ -50,7 +49,7 @@ export default class BuilderRegistry {
   resolveAmbiguousBuilders (builders, builderOverride) {
     function findBuilder (name) {
       const namePattern = new RegExp(`^${name}Builder$`, 'i')
-      return _.find(builders, builder => builder.name.match(namePattern))
+      return builders.find(builder => builder.name.match(namePattern))
     }
 
     let builder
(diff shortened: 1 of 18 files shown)


lib/composer.js
```diff
@@ -17,7 +17,7 @@ export default class Composer {
     this.configChange = atom.config.onDidChange('latex', () => {
       this.rebuildCompleted = new Set()
     })
-    this.checkEnvironment()
+    if (!atom.inSpecMode()) this.checkEnvironment()
```
thomasjo: Since this is not an "exit junction" (a `return` statement or similar), let's wrap this one in braces and split it across multiple lines.

spec/parsers/log-parser-spec.js
```diff
@@ -40,18 +40,18 @@ describe('LogParser', () => {
...
+      expect(_.countBy(result.messages, 'type').error).toBe(3)
     })
 
     it('associates an error with a file path, line number, and message', () => {
       const logFile = path.join(fixturesPath, 'errors.log')
       const texFile = path.join(fixturesPath, 'errors.tex')
       const parser = new LogParser(logFile, texFile)
       const result = parser.parse()
-      const error = _.find(result.messages, (message) => { return message.type === 'Error' })
+      const error = _.find(result.messages, (message) => { return message.type === 'error' })
```
thomasjo: Since we're changing this anyway, maybe we can change this to `Array.find`? But we can also do that in a separate PR at some point (remove as much lodash stuff as possible).

lib/builder-registry.js
```diff
@@ -17,6 +17,14 @@ export default class BuilderRegistry {
     return this.resolveAmbiguousBuilders(candidates, this.getBuilderFromMagic(filePath))
   }
 
+  async checkEnvironment () {
```
yitzchak: `checkRuntimeDependencies` sounds good.

PR #286, opened 2016-10-31
Add configuration check and migrator

yitzchak:
Move checks that were happening in `Composer` into `ConfigMigrator` to avoid race condition in the creation of `latex`.

### DIFF
--- a/lib/composer.js
+++ b/lib/composer.js
@@ -228,13 +228,6 @@ export default class Composer extends Disposable {
     return files
   }
 
-  async checkRuntime () {
-    latex.log.group('LaTeX Check')
-    await latex.builderRegistry.checkRuntimeDependencies()
-    latex.opener.checkRuntimeDependencies()
-    latex.log.groupEnd()
-  }
-
   moveResult (result, filePath) {
     const originalOutputFilePath = result.outputFilePath
     result.outputFilePath = this.alterParentPath(filePath, originalOutputFilePath)
--- a/lib/composer.js
+++ b/lib/composer.js
@@ -228,6 +228,13 @@ export default class Composer extends Disposable {
     return files
   }
 
+  async checkRuntime () {
+    latex.log.group('LaTeX Check')
+    await latex.builderRegistry.checkRuntimeDependencies()
+    latex.opener.checkRuntimeDependencies()
+    latex.log.groupEnd()
+  }
+
   moveResult (result, filePath) {
     const originalOutputFilePath = result.outputFilePath
     result.outputFilePath = this.alterParentPath(filePath, originalOutputFilePath)
--- a/lib/composer.js
+++ b/lib/composer.js
@@ -18,9 +18,6 @@ export default class Composer extends Disposable {
     this.disposables.add(atom.config.onDidChange('latex', () => {
       this.rebuildCompleted = new Set()
     }))
-    if (!atom.inSpecMode()) {
-      this.checkEnvironment()
-    }
   }
 
   initializeBuild (filePath) {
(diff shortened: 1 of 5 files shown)


lib/config-migrator.js
```diff
@@ -0,0 +1,61 @@
+/** @babel */
+
+import _ from 'lodash'
+import { heredoc } from './werkzeug'
+
+export default class ConfigMigrator {
```
thomasjo: Export a function instead? `migrateConfiguration`, or maybe even just `migrate`?
Then we can avoid the static functions being exported etc.

lib/config-migrator.js
```diff
@@ -0,0 +1,59 @@
+/** @babel */
+
+import _ from 'lodash'
+import { heredoc } from './werkzeug'
+
+export default async function checkConfigAndMigrate () {
```
thomasjo: Can we avoid having this be async? I'm just worried since we're calling it in a _fire and forget_ manner from the main module, and that introduces the possibility of a race condition. Either we make it sync (and wait for the async call to `latex.builderRegistry.checkRuntimeDependencies()`), or we have to wait in the main module.

yitzchak: I think that makes sense. It will have to with `await` since the builders are async.

thomasjo: Async in JS is implemented using Promises. You can `.then` from the sync side instead.

thomasjo: But of course, if making `activate` async works without side-effects, then we might as well just do that I guess. I really think we need to test that assumption though. AFAIK package activation was never designed to be async.

thomasjo:
I think we should strive to make minimal changes here, so safest bet is likely to exploit the fact that async calls are implemented using promises; hence revert the async-ification of the module activation, etc. This is in essense a bugfix, so let's focus on getting this shipped?

yitzchak:
Sounds good. Just looked at the Atom code for verification and it looks like it assumes `activate` is synchronous. Maybe push the `Promise` stuff out `bootstrap` so `main.js` can finish wiring up the system then call the `Promise` stuff last in `activate`?

lib/config-migrator.js
thomasjo: Now that I think about it, why are we even checking runtime dependencies here? This is supposed to be config related, yeah?

Runtime checks should be can be executed as a separate step. They can be async (but we need to handle resolution/rejection).

lib/config-migrator.js
```diff
@@ -0,0 +1,59 @@
...
+  latex.log.group('LaTeX Check')
```
thomasjo: Let's not log anything unless there's an error? If everything works as intended, and we have not migrated any configuration, we should be silent IMHO. The stuff we are doing here is of no relevance to users unless we actually had to migrate something. I think it's questionable even in that scenario whether we need to inform them.

yitzchak:
What about splitting the "Check runtime" stuff into a `latex:check-runtime` command? This would remove all of the logging and async stuff from here.

thomasjo:
❤️ it!

yitzchak:
:cool:

yitzchak:
Okay, everything looks good in CI, etc. I am planning on doing a patch release shortly after merge.

lib/composer.js
```diff
@@ -231,6 +228,13 @@ export default class Composer extends Disposable {
     return files
   }
 
+  async checkRuntime () {
```
thomasjo: Not a huge thing, but maybe we can move this to the `main` module?

yitzchak: Sure.

PR #288, opened 2016-11-01
Use more ES6/7 features

yitzchak:
Decrease dependence on lodash by using more ES6/7 features

- [x] `Array.filter`
- [x] `Array.find`
- [x] `Array.forEach`
- [x] `Array.includes`
- [x] `Array.map`
- [x] `Array.reduce`
- [x] `Array.some`
- [x] `Object.assign`
- [x] More use of object destructuring including rest

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -21,7 +21,7 @@ export default class Logger {
   }
 
   showMessage (message) {
-    message = _.assign({ timestamp: Date.now() }, _.pickBy(message))
+    message = Object.assign({ timestamp: Date.now() }, _.pickBy(message))
     if (this._group) {
       this._group.push(message)
     } else {
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -46,7 +46,7 @@ export default class Logger {
     const loggingLevel = atom.config.get('latex.loggingLevel')
     const showBuildWarning = loggingLevel !== 'error'
     const showBuildInfo = loggingLevel === 'info'
-    const filteredMessages = _.filter(this.messages, (message) => {
+    const filteredMessages = this.messages.filter(message => {
       return message.type === 'error' || (showBuildWarning && message.type === 'warning') || (showBuildInfo && message.type === 'info')
     })
 
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -53,7 +53,7 @@ export default class Logger {
   }
 
   static getMostSevereType (messages) {
-    return _.reduce(messages, (type, message) => {
+    return messages.reduce((type, message) => {
       if (type === 'error' || message.type === 'error') return 'error'
       if (type === 'warning' || message.type === 'warning') return 'warning'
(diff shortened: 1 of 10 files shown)


lib/loggers/default-logger.js
```diff
@@ -53,7 +52,7 @@ export default class DefaultLogger extends Logger {
   showErrorMarkersInEditor (editor, messages) {
     const filePath = editor.getPath()
     if (filePath) {
-      const m = _.filter(messages, message => {
+      const m = messages.filter(message => {
         return message.filePath && message.range && filePath.includes(message.filePath)
```
thomasjo: Let's nuke the braces so we can nuke the `return`.

lib/logger.js
```diff
@@ -46,7 +46,7 @@ export default class Logger {
     const loggingLevel = atom.config.get('latex.loggingLevel')
     const showBuildWarning = loggingLevel !== 'error'
     const showBuildInfo = loggingLevel === 'info'
-    const filteredMessages = _.filter(this.messages, (message) => {
+    const filteredMessages = this.messages.filter(message => {
       return message.type === 'error' || (showBuildWarning && message.type === 'warning') || (showBuildInfo && message.type === 'info')
```
thomasjo: Remove braces + return

lib/loggers/default-logger.js
```diff
@@ -53,9 +52,8 @@ export default class DefaultLogger extends Logger {
   showErrorMarkersInEditor (editor, messages) {
     const filePath = editor.getPath()
     if (filePath) {
-      const m = _.filter(messages, message => {
-        return message.filePath && message.range && filePath.includes(message.filePath)
-      })
+      const m = messages.filter(message =>
```
thomasjo: Rename `m` as `marker`?

yitzchak: Sounds good.

thomasjo: Wow, less than a minute response time — you're a 🤖, hehe!

yitzchak: Ha! It's because I am working on that branch right now. More commits soon!!!

yitzchak:
What about changing `heredoc` from

```javascript
heredoc (input) {
  if (input === null) { return null }

  const lines = _.dropWhile(input.split(/\r\n|\n|\r/), line => line.length === 0)
  const indentLength = _.takeWhile(lines[0], char => char === ' ').length
  const truncatedLines = lines.map(line => line.slice(indentLength))

  return truncatedLines.join('\n')
}
```

with something like

```javascript
heredoc (input) {
  if (input === null) { return null }

  const match = input.match(/^ +/m)
  if (match) {
    input = input.replace(new RegExp(`^${match[0]}`, 'mg'), '')
  }

  return input.trim()
}
```

Thereby removing lodash dependence there.

PR #291, opened 2016-11-02
Fix matching for relative output directory

### LINKED ISSUE #290: Output Directory cleaning does not work if directory is set outside of "project"
If I set OutputDirectory to "build": build works, opening the result works, clean (alt+crtl+c) works.

If I set OutputDirectory to "../build": build works, opening the result works, clean does not work (no file is deleted).

If I set OutputDirectory to "/some/absolute/path/build": build works but message parsing fails ("Parsing of log files failed", opening the result fails (no message), clean does not work (no file is deleted).

yitzchak:
- [x] Keep generated paths relative
- [x] Spec

Fixes #290.

### DIFF
--- a/lib/composer.js
+++ b/lib/composer.js
@@ -172,17 +172,17 @@ export default class Composer extends Disposable {
     const patterns = this.getCleanPatterns(builder, rootFilePath, jobname)
 
     for (const pattern of patterns) {
-      const absolutePattern = path.join(rootPath, pattern)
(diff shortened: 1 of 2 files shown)


thomasjo:
What do you think about adding some tests for this to prevent future regression? I guess ideally we had started with a failing test that became green after pushing the fix. But we might as well add one positive and one negative test, after the fact?

yitzchak:
I agree. The clean specs need to be filled out.

PR #292, opened 2016-11-03
Add qpdfview, Zathura and Windows shell open

yitzchak:
Add support for qpdfview and Zathura. Does not look for existence of the executable, instead it just tries to execute `qpdfview --help` or `zathura -v` since they should be in the path. This method could be used to make the `latex.okularPath` setting unneeded.

Add Windows shell open which functions the same as `xdg-open`, i.e. as a final fallback,

### DIFF
--- a/lib/openers/qpdfview-opener.js
+++ b/lib/openers/qpdfview-opener.js
@@ -18,13 +18,7 @@ export default class QPdfViewOpener extends Opener {
   }
 
   async open (filePath, texPath, lineNumber) {
-    const args = [
-      '--unique',
-      `"${filePath}#src:${texPath}:${lineNumber}:0"`
-    ]
-
-    const command = `${this.executable} ${args.join(' ')}`
-
+    const command = `${this.executable} --unique "${filePath}#src:${texPath}:${lineNumber}:0"`
     await latex.process.executeChildProcess(command, { showError: true })
   }
 
--- a/lib/openers/qpdfview-opener.js
+++ b/lib/openers/qpdfview-opener.js
@@ -1,6 +1,5 @@
 /** @babel */
 
-import _ from 'lodash'
 import Opener from '../opener'
 import { isPdfFile, isPsFile } from '../werkzeug'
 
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -230,6 +230,7 @@
       "order": 16
     },
     "opener": {
+      "Description": "Default opener to use when viewing PDF, PS or DVI files. `automatic` will try to find the viewer best suited to the file type and your current settings."
```
thomasjo: This needs to be `"description"`, also this package.json is invalid since this line is also missing a comma at the end.

lib/openers/qpdfview-opener.js
```diff
@@ -0,0 +1,35 @@
...
+    } catch (e) {}
+  }
+
+  async open (filePath, texPath, lineNumber) {
+    const command = `${this.executable} --unique "${filePath}#src:${texPath}:${lineNumber}:0"`
+    await latex.process.executeChildProcess(command, { showError: true })
+  }
+
+  async canOpen (filePath) {
+    if (_.isUndefined(this.exists)) {
```
thomasjo: Is there some special reason for using `_.isUndefined(this.exists)` instead of simply checking `this.exists === undefined`? Seems very unnecessary to retain a dependency on lodash for something like this.

---
For reference, my suggested replace above is exactly how `isUndefined` is [implemented](https://github.com/lodash/lodash/blob/4.16.4/lodash.js#L12138), a simple comparison to `undefined`.

package.json
```diff
@@ -230,16 +230,20 @@
       "order": 16
     },
     "opener": {
+      "Description": "Default opener to use when viewing PDF, PS or DVI files. `automatic` will try to find the viewer best suited to the file type and your current settings.",
```
thomasjo: Typo; needs to be `"description"`.

thomasjo:
The concerns I expressed in #295 apply here as well.

yitzchak:
We are waiting on qpdfview and Zathura until the need is more obvious. Windows shell open is split out in #300.

PR #295, opened 2016-11-06
Add support for Evince forks

yitzchak:
Add support for MATE's [Atril](https://github.com/mate-desktop/atril) and Cinnamon's [Xreader](https://github.com/linuxmint/xreader) forks of Evince. Since these forks use the same opening and SyncTeX logic just with different D-Bus names support is accomplished by sub-classing `EvinceOpener`. They do not support the `org.freedesktop.Application` interface used to activate the application, so open in background is probably not possible.

### DIFF
--- a/lib/openers/atril-opener.js
+++ b/lib/openers/atril-opener.js
@@ -17,4 +17,8 @@ export default class AtrilOpener extends EvinceOpener {
   constructor () {
     super(DBUS_NAMES)
   }
+
+  canOpenInBackground () {
+    return false
+  }
 }
--- a/lib/openers/atril-opener.js
+++ b/lib/openers/atril-opener.js
@@ -15,7 +15,7 @@ const DBUS_NAMES = {
 
 export default class AtrilOpener extends EvinceOpener {
   constructor () {
-    super(DBUS_NAMES)
+    super('Atril', DBUS_NAMES)
   }
 
   canOpenInBackground () {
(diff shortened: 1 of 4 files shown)


thomasjo:
Overall this is great, but I'm super concerned about the maintenance cost of this... we're adding support for a bunch of openers, but we have no automated tests for any of our openers. The result is that we're amassing a fairly big burden on ourselves to have a full suite of VMs with various operating system and applications combinations, and we need to keep manually testing to ensure everything is working. We still have to perform manual testing now, but the with fewer combinations, there's obviously significantly less effort.

yitzchak:
Although I did use a Mint VM to do a final test of Atril and Xreader, I did all of the development on my Arch install. Right now I have the following openers installed: pdf-view, Evince, Okular, Atril, Xreader, qpdfview and Zathura. I can switch back forth by just selecting using `latex.opener`.

That said, you make a very good point about tests. Definitely worth thinking about it and investigating.

thomasjo:
Testing on Arch is fine, but we really have to test on e.g. Ubuntu as well. I mostly wanted to raise this concern early, since this is something we really need to address in the not too distant future.

Once I get some more time, I can investigate automating some manner of opener integration spec suite that leverages e.g. Docker (and/or Packer) to automate the process. Not super important that we find a way of automating this as part of CI, but we really need to automate the actual testing process. Manual smoke testing doesn't scale if we keep adding more opener support (and it's easy to miss/forget stuff).

yitzchak:
:cool:

PR #228, opened 2016-09-21
Remove texify support

thomasjo:
This PR removes our `texify` support. Anyone wanting to continue to use the package after we merge this PR, will have to ensure they have `latexmk` installed.

---
- [x] Check for the presence of `latex.builder` in user config.
  - [x] Remove config value automatically?
  - [x] Present a user with a notification along the lines of the deprecation warning?

Issue #216, opened 2016-09-11
Improve test coverage


thomasjo:
This is a tracking issue to help us improve the overall test coverage.

Some functionality currently has little to no spec coverage, and this is something that needs to be addressed. The list below is updated as we discover (and resolve) functionality which needs better coverage.
- ~~`TexifyBuilder` has zero coverage, which **must** be improved.~~
- `KnitrBuilder` has OK coverage, but **should** be improved.
  - `canProcess`.
- `LatexmkBuilder` has OK coverage, but **should** be improved.
  - `canProcess`.
  - `logStatusCode`.
  - All permutations supported by `constructArgs`.
- `DefaultLogger` has zero coverage, which **must** be improved.
- `LogParser` has reasonable coverage, but **can** be improved.
- `BuilderRegistry` has poor coverage, which **must** be improved.
  - `getAllBuilders`.
  - ~~All permutations of `resolveAmbiguousBuilders`.~~
- `Composer` has less than ideal coverage, which **should** be improved.
  - Try to test `sync` — might need too much mocking...
  - `moveResult`.
  - `resolveOutputFilePath`.
  - `showResult`
  - Status bar indicator functionality.
- `Latex` has OK coverage, but the tests **should** be improved.

thomasjo:
This is being resolved in #224

yitzchak:
Resolved by #301 pending fresh review.

PR #301, opened 2016-11-12
Add unified build state

yitzchak:
This an experimental PR which adds a unified build state in an effort to decrease excessive magic comment parsing (currently parsed 9 times without root comment) and possibly move toward a build pipeline.

### Issues
- [x] Add specs
- [x] Clean up specs after specs are made functional
- [x] Move more build parameters like `enableSynctex` into `BuildState`
- [x] Remove `MasterTexFinder`


yitzchak:
@thomasjo Thoughts so far? There is a lot of work to do yet, but do you think this a direction worth going in?

thomasjo:
I'm a little bit uneasy with regards to the deeply nested object approach. I'm a fan of all external access and mutation of state being done via methods/functions. There are many advantages to this approach, but perhaps the most immediately obvious one is that it affords spying on stateful objects. Another advantage (which is connected to the first one) is that it makes it possible to use polymorphism/inheritance — makes it so much easier to override implementation.

Nested access patterns like you're using now — e.g. `state.results.fdb` — looks much nicer with value properties instead of `set` and `get` methods, but if you implement unit tests and try to spy on or mock out parts of the build state, you'll start to the feel the pain. I honestly think it's much safer to keep the externally visible access 1 level deep, regardless of how we persist it internally. I'm thinking the example above should be accessed via something along the lines of `state.getFdbResult()`.

Even if tests are not written first, I always keep in mind that I have to write them at some point, and that I might need to spy, stub, and mock stuff. That means I generally never use value properties for anything that is intended to be publicly accessed.

Also, I think including the `latex.json` stuff in this PR is too adding way too much scope to this PR. As I understand it, the intent of this PR was/is to unify the build state, not add brand new functionality. I think that aspect should be split into a separate PR which can then be reviewed and revised separately. It's a fairly important change by the way, which will need careful consideration and likely a fair bit of discussion. Once we settle on a format for that file, we can not easily change it, so it needs to be rock solid and future-proof. Even the name of the file is something that should be discussed in my opinion.

I'm still super-busy though, so I might not have time to discuss the above in any great detail in the next few weeks, but I'd rather postpone this (and hypothetical project file PR) for some weeks. 
_(That might already be your plan though.)_

yitzchak:
Sounds good. I'll split the PR.

I did not have the nested `results` object originally and just had a `state.fdb` property. I was thinking about a build pipeline when I moved it, but I think that should also be a separate PR. I'll move it back in the interim.

You are correct, I was planning on taking a while to let this one bake. Especially since it removes `texify` and `MasterTexFinder`. If we go down this route then some of the missing tests in #224 would probably have a different form. Maybe cherry-pick out of #224 or merge and rebase and leaving #216 open?

I'll probably have some follow-on questions/issues. I know you are busy right so no worries about a timetable.

yitzchak:
@thomasjo I think this PR is ready for a review whenever you have time. I have added some tests and have been running off of this branch for about two weeks with no issues. Please note that this PR is branched from #224 and includes #228 since some tests had to be updated for `BuildState` or `JobState`.

PR #310, opened 2016-11-24
Auto create output tree

### LINKED ISSUE #299: Output directory, sub-directories not considered
I have my tex files nested in several sub-folders, where each files uses the 'root' magic comment to refer to the master file; e.g. root = ../../thesis.txt
To keep my file structure clean, I use the output directory ".build", which is in the root directory of the project.
This does not compile, and I get the error "I can't write on file 'chapter1/content.aux'"
However, I found I can only compile, after I manually create a sub-folder named "chapter1" in the output directory.
Would it make sense to auto-create said sub-folders upon first compilation?

yitzchak:
Create sub-directories in output directory so LaTeX can create aux files for `\include` and friends. This resolves #299. Also removes deprecated `wrench` in favor of `fs-extra`.

### Remaining Issues
- [x] Specs
- [ ] Promisify file I/O?


yitzchak:
Closing. Will maintain branch in case we want to integrate this with non-DiCy builders.

PR #311, opened 2016-11-27
Add project file (Alpha Algorithm)

yitzchak:
This PR adds a project file versus magic comments. It is branched from #301 so I will just rebase as needed.

### Issues
- [x] What file format do we want? Currently YAML is used.
- [ ] What naming pattern do we want? Currently `latex.yaml` is used.
- [ ] Search and property application algorithm? Currently *Alpha* is used.
- [ ] Specs

### Alpha Algorithm 
1. Walk upward from the directory of the current file until the parent project folder is reached.
2. If `latex.yaml` is found then parse and apply it.
  a. Apply the properties found in the root of the parsed file including looking for a `root` field.
  b. Look for file specific section. If `root` field is found then apply the properties of the section with that name, otherwise apply the current section.
3. Project file was not found so parse and apply the magic comments. 

#### Alpha Examples

Single file using only the 'root' section:

```yaml
enableExtendedBuildMode: false
engine: xelatex
jobNames:
  - foo
  - bar
```

Multiple subfiles with a single `root` document:

```yaml
root: wibble.tex
engine: uplatex
producer: dvips
```

Multiple 'root' documents sharing common properties:

```yaml
enableShellEscape: true
outputDirectory: build
foo/bar.tex:
  root: wibble.tex
wibble.tex:
  program: lualatex
snafu.tex:
  jobNames:
    - foo
    - bar
```

### Beta Algorithm

1. Parse and apply the magic properties. This includes looking for the `root` comment.
2. If the name of the root document is `quux.tex` then look for `quux.yaml` or `quux.yml`.
3. If the project file is found then parse and apply it. Note that any `root` field in the project file is not used and only a single level project file is needed.

## Beta Examples

All project files are single level:
```yaml
cleanPatterns:
  - '**/*.foo'
  - /wibble
enableSynctex: false
```


thomasjo:
Really, really glad to see YAML instead of JSON 😍

yitzchak:
:cool: I agree. YAML just makes more sense.

yitzchak:
The Beta algorithm is over on the [add-project-file-beta](https://github.com/thomasjo/atom-latex/tree/add-project-file-beta) branch for reference.

yitzchak:
Closing as Beta is preferred.

PR #312, opened 2016-11-29
Add build settings file

### LINKED ISSUE #305: Config file in addition to magic comments
Hi there, and thanks for all the hard work on this very useful extension.

I was wondering, would it be possible to set up a configuration mechanism using a config file (for instance a .latex.json inside the folder) in addition to magic comments. I know it is a pain to implement but this would be a very good thing for two reasons:

- magic comments are verbose and take space inside the latex documents this is boring, especially for users using editors not compatible with them,
- magic comments are necessarily shared as they are located in the tex files, but it is very common that users want a special behavior for a particular machine therefore a seprated config file would be better suited.

So please consider making this possible ;)

Best regards!

yitzchak:
This PR has an alternate algorithm for build settings files in #311. Resolves #305

### Issues
- [x] Naming/extension. *We will use `.yaml`*
- [x] Specs

### Beta Algorithm

1. Parse and apply the magic properties. This includes looking for the `root` comment.
2. If the name of the root document is `quux.tex` then look for `quux.yaml`.
3. If the build settings file is found then parse and apply it. Note that any `root` field in the build settings file is not used and only a single level settings file is needed.

## Examples

All settings files are single level:
```yaml
cleanPatterns:
  - '**/*.foo'
  - /wibble
enableSynctex: false
```


lib/composer.js
```diff
@@ -20,45 +22,145 @@ export default class Composer extends Disposable {
...
+
+  getMagic (state) {
+    return new MagicParser(state.getFilePath()).parse()
+  }
+
+  initializeBuildStateFromProjectFile (state) {
+    try {
+      const { dir, name } = path.parse(state.getFilePath())
+      const filePattern = path.format({ dir, name, ext: '.@(yaml|yml)' })
+      const filePath = _.first(glob.sync(filePattern))
```
thomasjo: Let's just pick one extension and avoid the globbing?

yitzchak: Sounds fine. Do you have a preference?

thomasjo: Since the official recommendation is to use `.yaml`, let's pick that? I don't really care either way, but there is no reason we need to support both extensions.

yitzchak: :+1:

yitzchak:
@thomasjo Are we satisfied with this PR? I know you wanted to be cautious since it will be difficult to change later on.

Also, I changed the terminology from "Project File" to "Build Settings File".

PR #325, opened 2017-01-01
Add Zathura Opener

### LINKED ISSUE #324: Add support for zathura viewer
Please add the support to `zathura` viewer under linux.

yitzchak:
Resolves #324.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -69,7 +69,7 @@ export default {
   async checkRuntime () {
     latex.log.group('LaTeX Check')
     await latex.builderRegistry.checkRuntimeDependencies()
-    latex.opener.checkRuntimeDependencies()
+    await latex.opener.checkRuntimeDependencies()
     latex.log.groupEnd()
   }
 }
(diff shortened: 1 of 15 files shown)


yitzchak:
I was trying to avoid adding another setting like `zathuraPath` to our settings page.

Since Zathura is on Linux it should be in the search path so `zathura -v` can be used to verify its existence. That makes `initialize` async if we use `ProcessManager`. We can't call the initialize function in the constructor because it uses the `latex` global which isn't defined yet, so that makes `canOpen` async also.

As a side note having `canOpen` async would probably help `EvinceOpener` avoid the very tiny possibility of having not completed it's initialization before `canOpen` is called.

thomasjo:
Again, I might be missing something obvious, but how does making a function asynchronous avoid a race condition? Wouldn't it potentially just exacerbate the situation, or even worse, just hide the problem?

I think at this point it might make more sense to add the `zathuraPath` setting even though I would love to avoid it as well. I'm a little bit concerned about making this a special case though, and potentially forcing people to add it to their `PATH` environment variable (if it it isn't there for some reason).

We should probably refactor parts of the code soon anyway, to get rid of race conditions and unnecessary complexity.

If you feel strongly about this, and if you feel confident it *actually* fixes the race condition, go ahead and merge — we can always revert/refactor this later.

yitzchak:
Okay. I'll make it work like `EvinceOpener` and remove async from `canOpen`.

PR #329, opened 2017-01-04
WIP: Auto configure path

yitzchak:
This is an experimental PR to automatically configure the search path for the TeX distribution and viewers.


yitzchak:
Closing in light of significant code change. Will leave branch active in case we want to revisit at some point in the future.

PR #343, opened 2017-02-05
Opt-in automatic clean after build

oguzserbetci:
Hello,

I added a setting (off by default) to clean-up after builds. Let me know what you think of it. I would then write the tests after you give the ok :)

#### TODO
- [ ] Tests

--- 
PS: This has been my first time hacking an atom package. I'm not entirely sure if I've skipped something.

### DIFF
--- a/lib/composer.js
+++ b/lib/composer.js
@@ -192,10 +192,10 @@ export default class Composer extends Disposable {
 
     const jobs = state.getJobStates().map(jobState => this.buildJob(builder, jobState))
 
-    await Promise.all(jobs)
+    const statuses = await Promise.all(jobs)
 
-    if (this.shouldCleanAfterBuild()) {
-      this.clean()
+    if (statuses.every(status => status) && this.shouldCleanAfterBuild()) {
+      await this.clean(false)
     }
 
     latex.log.groupEnd()
@@ -221,7 +221,11 @@ export default class Composer extends Disposable {
       }
     } catch (error) {
       latex.log.error(error.message)
+
+      return false
     }
+
+    return true
   }
 
   async sync () {
@@ -248,7 +252,7 @@ export default class Composer extends Disposable {
     await latex.opener.open(outputFilePath, filePath, lineNumber)
   }
 
-  async clean () {
+  async clean (startGroup = true) {
     const { filePath } = getEditorDetails()
     if (!filePath || !this.isTexFile(filePath)) {
       return false
@@ -257,13 +261,13 @@ export default class Composer extends Disposable {
     const { builder, state } = this.initializeBuild(filePath, true)
     if (!builder) return false
 
-    latex.log.group('LaTeX Clean')
+    if (startGroup) latex.log.group('LaTeX Clean')
 
     const jobs = state.getJobStates().map(jobState => this.cleanJob(builder, jobState))
 
     await Promise.all(jobs)
 
-    latex.log.groupEnd()
+    if (startGroup) latex.log.groupEnd()
   }
 
   async cleanJob (builder, jobState) {
--- a/lib/composer.js
+++ b/lib/composer.js
@@ -194,6 +194,10 @@ export default class Composer extends Disposable {
 
     await Promise.all(jobs)
 
+    if (this.shouldCleanAfterBuild()) {
+      this.clean()
+    }
+
     latex.log.groupEnd()
   }
 
@@ -395,6 +399,8 @@ export default class Composer extends Disposable {
   shouldMoveResult (jobState) {
     return jobState.getMoveResultToSourceDirectory() && jobState.getOutputDirectory().length > 0
   }
+
+  shouldCleanAfterBuild () { return atom.config.get('latex.cleanAfterBuild') }
 
   shouldOpenResult () { return atom.config.get('latex.openResultAfterBuild') }
 }
--- a/lib/composer.js
+++ b/lib/composer.js
@@ -213,19 +213,20 @@ export default class Composer extends Disposable {
 
       if (statusCode > 0 || !jobState.getLogMessages() || !jobState.getOutputFilePath()) {
         this.showError(jobState)
-      } else {
-        if (this.shouldMoveResult(jobState)) {
-          this.moveResult(jobState)
-        }
-        this.showResult(jobState)
+        return false
       }
+
+      if (this.shouldMoveResult(jobState)) {
+        this.moveResult(jobState)
+      }
+      this.showResult(jobState)
+
+      return messages.every(message => message.type !== 'error')
     } catch (error) {
       latex.log.error(error.message)
-
-      return false
     }
 
-    return true
+    return false
   }
 
   async sync () {
--- a/package.json
+++ b/package.json
@@ -164,11 +164,17 @@
       ],
       "order": 8
     },
+    "cleanAfterBuild": {
+      "title": "Clean after a build.",
+      "type": "boolean",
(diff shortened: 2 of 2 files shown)


yitzchak:
Thank you for your PR! We will take a look at it as soon as we can.

package.json
```diff
@@ -164,11 +164,17 @@
       ],
       "order": 8
     },
+    "cleanAfterBuild": {
+      "title": "Clean after a build.",
```
thomasjo: Change the title to `"Clean after Build"`, and perhaps consider adding a description giving some additional details, such as the fact that cleanup (should) only take place after a *successful* build.

thomasjo (changes requested):
I'm pretty convinced we should only perform automatic cleanup if the build (meaning all jobs) finished successfully. If we perform a cleanup for a job that failed, it will be extremely cumbersome for a user to track down the errors since the log file containing the necessary information has been removed.

In summary: ensure the opt-in automatic cleanup is only executed if all jobs finished successfully.

@yitzchak Do you think we need to limit this based on the log level perhaps? If we nuke log files containing warnings, and the user wants to see them, this new behavior will cause issues.

lib/composer.js
```diff
@@ -194,6 +194,10 @@ export default class Composer extends Disposable {
 
     await Promise.all(jobs)
 
+    if (this.shouldCleanAfterBuild()) {
+      this.clean()
```
yitzchak: This needs to be `await this.clean()` since `clean` is `async`.

We probably need to add a boolean parameter to clean specifying whether `latex.log.group()` and `latex.log.groupEnd()` need to be called. Currently this call to clean will fry any warnings/errors generated by the build because of the embedded calls to to `latex.log.group`.

yitzchak:
@thomasjo Since all logs have been parsed by the time `clean` has been called I think we will be okay if they get nuked.

We do need to prevent `clean` from resetting the log by calling `latex.log.group` or `latex.log.groupEnd`. Maybe with a boolean parameter argument list of `clean`?

Also, `cleanExtensions`, `moveResult`, etc. are all now settings in `BuildState` so we might consider making this also.

lib/composer.js
```diff
@@ -194,6 +194,10 @@ export default class Composer extends Disposable {
 
     await Promise.all(jobs)
 
+    if (this.shouldCleanAfterBuild()) {
```
yitzchak: @oserbetci We are not currently tracking the success/failure of the build. We will need some extra code to make the clean conditional.

Since error messages could come from the log, not just from a returned status code (which will generate a log message), one option would be to add the following to `Logger`.

```javascript
hasPendingErrors () {
  return this._group && this._group.some(message => message.type === 'error')
}
```

Then you could do:

```javascript
if (!latex.log.hasPendingErrors() && this.shouldCleanAfterBuild()) {
```

thomasjo: Chucking this onto a logger feels like a hack. A logging facility shouldn't be responsible for keeping track of build success/error state. If this is currently the best way of solving this, we're going to have to do some refactoring.

yitzchak: @thomasjo We could also have `buildJob` return false if an error is thrown or if an error is sent to the log:

```javascript
  async buildJob (builder, jobState) {
    try {
      const statusCode = await builder.run(jobState)
      builder.parseLogAndFdbFiles(jobState)

      const messages = jobState.getLogMessages() || []
      for (const message of messages) {
        latex.log.showMessage(message)
      }

      if (statusCode > 0 || !jobState.getLogMessages() || !jobState.getOutputFilePath()) {
        this.showError(jobState)
        return false
      }

      if (this.shouldMoveResult(jobState)) {
        this.moveResult(jobState)
      }
      this.showResult(jobState)

      return messages.every(message => message.type !== 'error')
    } catch (error) {
      latex.log.error(error.message)
    }

    return false
  }
```

Throughts?

lib/composer.js
oguzserbetci: Thanks for pointing this out. Excuse me for the mistakes I do as I try to grasp Javascript on the way 🙈

package.json
oguzserbetci: I will push a description with the title fix. Let me know if you don't like the wording :)

lib/composer.js
oguzserbetci: I'm not entirely sure how everything works here but IMO returning false _feels_ better.

lib/composer.js
yitzchak: No worries! We appreciate the help.

package.json
```diff
@@ -165,7 +165,8 @@
       "order": 8
     },
     "cleanAfterBuild": {
-      "title": "Clean after a build.",
+      "title": "Clean after build.",
```
thomasjo: Change this to `"title": "Clean after Build"`

lib/composer.js
yitzchak: Starting at line 195, I suggest:

```javascript
const statuses = await Promise.all(jobs)

if (statuses.every(status => status) && this.shouldCleanAfterBuild()) {
  this.clean(false)
}
```

Change the signature of `clean` to `async clean (startGroup = true)` and put a guard conditional before the log `group`/`groupEnd` calls. For instance,

```javascript
if (startGroup) latex.log.group('LaTeX Clean')
```

yitzchak:
I have checked out a copy and it appears to work. I added one comment regarding the setting description, but aside from that I think it is time to write some tests.

package.json
```diff
@@ -164,11 +164,18 @@
       ],
       "order": 8
     },
+    "cleanAfterBuild": {
+      "title": "Clean after Build",
+      "description": "Run clean command after every successful build.",
```
yitzchak: The default settings for `cleanPatterns` perform a deep clean, i.e. delete all DVI, PS, PDF and synctex.gz files. If the user enables `cleanAfterBuild` without changing the default patterns then the build will execute and everything will be deleted.

At a minimum we need some language here explaining that those extensions should probably be removed from `cleanPatterns`. At some point we might consider checking the pattern list explicitly, but it seems to me we would need the explanation here in any case.

oguzserbetci: How is this wording?
> Run clean command after every successful build. **Note:** Default clean patterns includes files that you might want to keep, e.g. PDF, DVI and PS. Make sure to have a look at the default patterns before turning this setting on.

I can't do a new line :/ and the bold is also not much more readable.

screen shot 2017-03-01 at 18 17 25

yitzchak: I suggest a couple of tweaks.

> Run clean command after every successful build. Please note that the default `Clean Patterns` setting includes output files such as PDF, DVI, PS and SyncTeX files. These should be removed before turning this setting on.

thomasjo: I'm not overly eager on adding a feature that introduces a caveat like this. Instead of trying to "fix" the problem by adding this description, perhaps we should take a step back and look at the broader picture — e.g. would it be better if we changed the default clean pattern to not include result and SyncTeX files? Or perhaps there is some other alternative, that is better than a note in a description (that frankly put, I fear many will overlook/ignore).

**TL; DR** instead of trying to fix the symptoms, fix the cause of the symptoms instead :thought_balloon:

yitzchak: One option would be to remove the extensions from `cleanPatterns` thereby making `latex:clean` essentially a shallow clean. We could add a `latex:deep-clean` command which then included them. The "Clean After Build" would then be equivalent to `latex:build` -> `latex:clean` not `latex:build` -> `latex:deep-clean`.

PR #355, opened 2017-03-24
Refocus Atom on Evince Reverse Sync

### LINKED ISSUE #354: Evince backward sync doesn't focus Atom window
Unfortunately I experience another small annoyance. Backward sync from Evince (Crtl+click) does highlight the corresponding line in Atom but the focus stays on Evince window. 

- Do I need to change some setting in Evince?
- What other information could be helpful?

yitzchak:
Add window focus on reverse sync for Evince.

Resolves #354

### DIFF
--- a/lib/openers/evince-opener.js
+++ b/lib/openers/evince-opener.js
@@ -20,6 +20,7 @@ const DBUS_NAMES = {
 function syncSource (uri, point) {
   const filePath = decodeURI(url.parse(uri).pathname)
   atom.workspace.open(filePath).then(editor => editor.setCursorBufferPosition(point))
+  atom.focus()
 }
 
 export default class EvinceOpener extends Opener {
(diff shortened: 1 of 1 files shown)


lib/openers/evince-opener.js
```diff
@@ -19,8 +19,8 @@ const DBUS_NAMES = {
 
 function syncSource (uri, point) {
   const filePath = decodeURI(url.parse(uri).pathname)
-  atom.workspace.open(filePath).then(editor => editor.setCursorBufferPosition(point))
   atom.focus()
+  atom.workspace.open(filePath).then(editor => editor.setCursorBufferPosition(point))
```
dmohns: Why do you focus first and then sync? I tried out both and either seems to work. But me personally feel that syncing first and then focus feels a bit more natural. Is there a technical motivation behind it?
(In the case of first focus and sync I could sometimes see the content of my editor windows move around.)

PS: This is so awesome.

yitzchak: It probably doesn't really matter. Focusing first gives some visual cues to the movement of the cursor, though.

PR #364, opened 2017-04-29
WIP: Add Dicy builder

yitzchak:
Add support for the [Dicy](https://yitzchak.github.io/dicy/) (Directed Cyclic graph) builder.

- [x] Add fallback for `latex:deep-clean`
- [x] Add clean after build option
- [ ] Add specs


yitzchak:
@thomasjo Can you give me a review on this one? I haven't written specs yet so I think it just needs a "sanity" check right now.

thomasjo:
I'll have a look at this tonight. Really excited about this!

package.json
```diff
@@ -133,42 +141,27 @@
...
-        "**/*.nav",
-        "**/*.out",
-        "**/*.pdf",
-        "**/*.ps",
-        "**/*.snm",
-        "**/*.synctex.gz",
-        "**/*.toc",
-        "/**/_minted-{jobname}",
-        "/{output_dir}/sage-plots-for-{jobname}.tex",
+        "$OUTDIR/$JOB!($OUTEXT|.synctex.gz|.tex)",
```
thomasjo: Why are we changing the defaults here? Seems like this breaks current functionality.

yitzchak: Dicy has separate clean levels (shallow/deep) so this was changed to make that work. I'll detail more below.

yitzchak:
**TL;DR** I'll split into two PRs unless you feel otherwise.

Dicy has some build modes that resolve #343 right out of the box. Basically there is the usual build mode

```javascript
dicy.run('load', 'build', 'log', 'save')
```

which is equivalent to our current `latex:build` command (with log parsing) and 

```javascript
dicy.run('load', 'clean', 'save')
```

which is our `latex:clean` command. The commands can be composed in any order to create a "clean after build" command

```javascript
dicy.run('load', 'build', 'log', 'clean', 'save')
```

In order to make this command not delete the output targets produced by the `build` command I added a `deepClean` option which makes the regular `clean` command more of a shallow clean. That means that there is no reason to include the `*.pdf` type patterns in the `cleanPatterns`.  So shallow clean (remove everything but PDF, DVI, etc) will be done when `clean` is passed to Dicy unless the following is done

```javascript
dicy.run('load', 'clean', 'save', { deepClean: true })
```

Hence adding `latex:deep-clean` command along with the stuff from #343. Hope that explains my logic a bit. Again, I really don't have a problem splitting into two PRs, maybe a "minimal Dicy integration" and a "full Dicy integration?" No one is really using Dicy yet, so we can also change the logic it is using if you see a better way to resolve this issue.

yitzchak:
Closed in favor of #367.

PR #367, opened 2017-05-09
Add basic support for DiCy builder

yitzchak:
Add support for [DiCy](https://yitzchak.github.io/dicy/) without changing current behavior of atom-latex.

- [x] Basic support for DiCy
- [x] Add `latex:kill` support
- [x] Add `texPath` support
- [x] Specs

### DIFF
--- a/lib/composer.js
+++ b/lib/composer.js
@@ -3,7 +3,8 @@
 import _ from 'lodash'
 import fs from 'fs-plus'
 import path from 'path'
-import { getEditorDetails, replacePropertiesInString } from './werkzeug'
+import { Dicy } from '@dicy/core'
+import { getEditorDetails, isDviFile, isPdfFile, isPsFile, replacePropertiesInString } from './werkzeug'
 import minimatch from 'minimatch'
 import glob from 'glob'
 import yaml from 'js-yaml'
@@ -14,11 +15,16 @@ import MagicParser from './parsers/magic-parser'
 export default class Composer extends Disposable {
   disposables = new CompositeDisposable()
   cachedBuildStates = new Map()
+  cachedDicy = new Map()
 
   constructor () {
     super(() => this.disposables.dispose())
     this.disposables.add(atom.config.onDidChange('latex', () => {
       this.rebuildCompleted = new Set()
+      const dicy = this.cachedDicy.values().next().value
+      if (dicy) {
+        dicy.updateOptions(this.getDicyOptions(), true)
+      }
     }))
   }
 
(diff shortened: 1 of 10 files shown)


yitzchak:
@thomasjo :tada: This one should be ready for a review. :eyes:

lib/composer.js
```diff
@@ -162,6 +168,102 @@ export default class Composer extends Disposable {
...
+      const typeText = event.category ? `${event.category}: ` : ''
+      const message = {
+        type: event.severity,
+        text: `${nameText}${typeText}${event.text.replace('\n', ' ')}`
+      }
+
+      if (event.source) {
+        message.filePath = path.resolve(dicy.rootPath, event.source.file)
+        if (event.source.start) {
+          message.range = [[event.source.start - 1, 0], [(event.source.end || event.source.start) - 1, 65535]]
```
thomasjo: Extract these "magic numbers" — 65535 — into a constant? Perhaps `MAX_LINE_LENGTH` or whatever it represents. The value seems a bit arbitrary, perhaps consider `Number.MAX_SAFE_INTEGER` instead, unless there is a specific reason for the 64k limit.

yitzchak: Good point. I think I used the same silliness in `LogParser` so it will have to fixed there also.

yitzchak:
@thomasjo I think this one is ready for merge. Thoughts?

lib/composer.js
```diff
@@ -14,11 +15,16 @@ import MagicParser from './parsers/magic-parser'
...
+  cachedDicy = new Map()
 
   constructor () {
     super(() => this.disposables.dispose())
     this.disposables.add(atom.config.onDidChange('latex', () => {
       this.rebuildCompleted = new Set()
+      const dicy = this.cachedDicy.values().next().value
+      if (dicy) {
+        dicy.updateOptions(this.getDicyOptions(), true)
+      }
```
thomasjo: What exactly is this the purpose of this section? Seems like this is somewhat non-deterministic. If you only grab the first instance, and update the options on that one, the rest of them are left untouched right? And whatever is first, is entirely dependent upon the what file you build first?

yitzchak: `updateOptions` writes the options by writing them into `.dicy.yaml` in the user's home directory. All  builder instances will notice the change in timestamp/hash and update their options in response. So it doesn't really matter which one it get's called on.

Conceptually, it should probably be a static member function, but I think I had to call some instance functions in it.

thomasjo: OK, that is not at all clear from the code, so I think this logic needs to be extracted into a function with a descriptive name. Or at least this needs a comment documenting that behavior.

yitzchak: Sound good. Will do.

yitzchak:
I think I've got the remaining issues resolved on this one. Can you give it a look again?

package.json
```diff
@@ -116,11 +117,17 @@
       "default": true,
       "order": 5
     },
+    "enableDicy": {
+      "description": "Enable the usage of the [DiCy](https://yitzchak.github.io/dicy/) builder instead of `latexmk`.",
+      "type": "boolean",
```
yitzchak: @thomasjo We could use an enumeration here and call the setting `builder` again?

thomasjo: @yitzchak Yeah, I think I like that idea. Unless of course we envision moving entirely to DiCy in the (not too distant) future.

yitzchak: @thomasjo Cool. I'll make the change since I don't think it would prevent a shift to DiCy if we decide to go there.

Aside from that I think it should be ready to :ship:

thomasjo: My point was that if we think we'll migrate entirely to DiCy at some point not too far into the future, it is more user friendly to have an experimental "Enable DiCy" setting that will get removed once we migrate. If we decide to keep both latexmk and DiCy around for a while, down the line, we can always re-introduce the "Builder" enumeration setting then.

Unless you disagree, or have a different viewpoint, I think we should stick with what you originally had here. Also think we should call it "Use DiCy builder" or some such, and it might make sense to point out that this is an experimental feature for the time being. Thoughts?

yitzchak: Ok, you have a good point. I'll roll it back and rename it.

yitzchak:
@thomasjo I've rolled back the setting changes. Can you check the option description (or anything else that comes to mind) when you get a chance?

thomasjo:
:ship:

Issue #359, opened 2017-04-17
Feature request: filter of LaTeX Log


tririver:
Hi, thomasjo,

Thank you for the great package! This package is my favorite in Atom, and is what made it possible for me to convert from Emacs to Atom :)

The "LaTeX Log" feature is among one of my favorite features. However, it would be nicer if in the settings, one can use regexp (or something like that) to blacklist some warning messages. For example, the below is what my typical messages look like. On the other hand, I still hope to see some warning messages, for example, missing references, etc. So I don't like to change "Logging Level" to error.

![latex_warning](https://cloud.githubusercontent.com/assets/1221231/25076199/1ec8be5c-2351-11e7-900f-5e652aae7230.png)

Best regards,
Yi

yitzchak:
Thank you for the feature request! We will look into it.

PR #372, opened 2017-05-16
Improve Log Message Reporting

yitzchak:
This a PR to improve the log reporting UI. Various ideas are listed below.

- [x] Use the new Docks API
- [x] Use events to make message display real-time (not cached until end of build)
- [x] Add message count indicator
- [x] Update busy indicator and fix the horrible wobble of the icon
- [x] Refresh log indicators on `loggingLevel` setting change
- [x] Use table for message display
- [x] Mark error locations in log files
- [x] ~~Allow logging level to be set from the dock~~ Move to filtering UI
- [x] ~~Add message copy~~ Move to separate PR
- [x] ~~Add message filtering UI (#359)~~ Move to separate PR
- [x] ~~Retain messages from previous builds~~ requires changes to log message format
- [x] Serialize messages


yitzchak:
@thomasjo Check the attached screen capture showing usage of the Docks API, message counter indicator and updating the log in response to changes in `loggingLevel`.
screen capture

yitzchak:
:fire: wobble in busy indicator.

peek 2017-05-19 08-09

yitzchak:
Table based display of messages.

screenshot from 2017-05-21 11-37-31

yitzchak:
Screenshot showing `loggingLevel` controls in dock.

peek 2017-05-23 09-25

thomasjo:
> Screenshot showing loggingLevel controls in dock.

I like the intent, but from a user experience perspective it's not clear what the buttons do, nor what actually happens when you click them. Maybe a tooltip is enough, but this likely needs a label or something near the buttons that conveys "Adjust logging level". Couple that with a descriptive tooltip that clearly states what logging level here means.

What do you think? :thought_balloon:

yitzchak:
Agree on log filter UI. Will probably need multiple iterations so I am moving into separate PR.

yitzchak:
@thomasjo I've removed the logging level buttons and done some code cleanup. I think this one should ready to go.

lib/views/message-count.js
```diff
@@ -0,0 +1,44 @@
...
+  async destroy () {
+    await etch.destroy(this)
+    this.disposables.dispose()
+  }
+
+  render () {
+    const loggingLevel = atom.config.get('latex.loggingLevel')
+
+    if ((loggingLevel === 'error' && this.properties.type !== 'error') ||
+      (loggingLevel === 'warning' && this.properties.type === 'info')) {
```
thomasjo: Maybe we can extract two booleans here with descriptive names to make it a little bit clearer what we're checking?

lib/views/status-label.js
```diff
@@ -17,33 +18,23 @@ export default class StatusLabel {
...
-    if (this.properties.type) classNames.push(`latex-${this.properties.type}`)
-    if (this.properties.onClick) classNames.push('latex-status-link')
-
-    let iconClassNames = [
-      'icon',
-      `icon-${this.properties.icon}`
-    ]
-
-    if (this.properties.spin) iconClassNames.push('latex-spin')
+    if (this.properties.busy) classes.push('is-busy')
```
thomasjo: Pushing to a constant; this might break at some point in the future.

lib/views/status-label.js
```diff
@@ -17,33 +18,23 @@ export default class StatusLabel {
...
 
     return (
-      <div className={classNames.join(' ')} onclick={this.properties.onClick}>
-        {this.properties.icon ? <div className={iconClassNames.join(' ')} /> : ''}
-        <span>{this.properties.text}</span>
+      <div className={classes.join(' ')} onclick={() => latex.log.show()}>
```
thomasjo: Perhaps introduce a method along the lines of `getClassNames()` or some such. That would also circumvent the problem above, altering a constant array.

yitzchak:
The test coverage went down by one because `getMostSevereType` is no longer being used. I used CSS specificity instead to make the error markers in the gutter override the warning markers, etc. The advantage of this method is that if two message ranges overlap then the non-overlapping regions will still be styled correctly.

Aside from that, I added a few more tests and fixed the issues you noted. Once I get a green light from CI I'll probably pull the trigger! :tada:

yitzchak:
@thomasjo I've managed to recover most of the tests that are still applicable. I'll take a look at specs for `MarkerManager` tomorrow. I think that will wrap it up.

yitzchak:
Using Atom's loading indicator instead?

peek 2017-06-01 09-24

PR #387, opened 2017-06-16
Adjust to upstream async API changes

### LINKED ISSUE #385: Changes required due to TextBuffer.save becoming async in Atom 1.19
Hey @thomasjo!

In Atom v1.19, we will release a [major change](https://github.com/atom/atom/pull/14435) to Atom's core text buffer data structure. As part of this change, we have made `TextBuffer.save` asynchronous; rather than blocking until the save is complete, it now immediately returns a `Promise` that *resolves* when the save is complete. Because of this, a few other Atom APIs that *use* `save` have similarly become async:

* `Pane.close`
* `TextBuffer.save`
* `TextEditor.save`
* `Pane.saveItem`
* `Pane.saveItemAs`
* `Pane.saveActiveItem`
* `Pane.saveActiveItemAs`
* `Pane.saveItems`
* `Workspace.saveActivePaneItem`
* `Workspace.saveActivePaneItemAs`

#### Effects on this package

We think this package could be impacted by this upgrade because it calls the changed methods in the following places:

* `TextEditor.save`
  * [here](https://github.com/thomasjo/atom-latex/tree/master/lib/composer.js#L176)

We found these calls using a regex search, so this list might be incomplete, and it might contain some false positives.

#### What to do about the change

It should be pretty easy to adjust your package code and/or tests to work with the new async behavior, and to simultaneously keep it working with older versions of Atom. Here are some examples of pull requests we opened on our bundled packages to cope with the change:

* https://github.com/atom/autocomplete-plus/pull/852
* https://github.com/atom/whitespace/pull/155
* https://github.com/atom/symbols-view/pull/222
* https://github.com/atom/status-bar/pull/193
* https://github.com/atom/tabs/pull/448

Please let me know if you have any questions. I would be happy to help!

thomasjo:
Atom's `TextEditor.save` is now async, and this PR addresses that change.
Since the change will land in Atom v1.19, the engine requirement was also updated accordingly.

Resolves #385

### DIFF
--- a/lib/composer.js
+++ b/lib/composer.js
@@ -173,7 +173,7 @@ export default class Composer extends Disposable {
     }
 
     if (editor.isModified()) {
-      editor.save() // TODO: Make this configurable?
+      await editor.save() // TODO: Make this configurable?
     }
 
     const { builder, state } = this.initializeBuild(filePath)
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     ]
   },
   "engines": {
-    "atom": ">=1.17.0 <2.0.0"
+    "atom": ">=1.19.0 <2.0.0"
   },
   "activationCommands": {
     "atom-workspace": [


thomasjo:
@yitzchak Let's ship this?

yitzchak:
@thomasjo I agree.

PR #398, opened 2017-08-12
Fix pdf-view opener splitting

### LINKED ISSUE #397: Multiple job names opens multiple right panes with pdf-view
Currently we try to open the pdf in a right pane with pdf-view. When there are multiple job names syncing or building causes a new right pane to be created for every job name versus creating a new tab in the same pane.

The following screenshot is with seven job names.

screenshot from 2017-08-09 12-00-11

yitzchak:
Prevent multiple splitting of pdf-view and add `pdfViewSplitDirection` configuration. Fixes #397.

### DIFF
--- a/lib/openers/pdf-view-opener.js
+++ b/lib/openers/pdf-view-opener.js
@@ -3,10 +3,6 @@
 import Opener from '../opener'
 import { isPdfFile } from '../werkzeug'
 
-function sleep (time) {
-  return new Promise(resolve => setTimeout(resolve, time))
-}
-
 export default class PdfViewOpener extends Opener {
   async open (filePath, texPath, lineNumber) {
     const texPane = atom.workspace.paneForURI(texPath)
@@ -22,14 +18,8 @@ export default class PdfViewOpener extends Opener {
     }
 
     const item = await atom.workspace.open(filePath, options)
-    const pane = atom.workspace.paneForItem(item)
-    if (pane && item && item.forwardSync) {
+    if (item && item.forwardSync) {
       item.forwardSync(texPath, lineNumber)
-      // If we don't wait for pdf-view to do it's first update then it will just
-      // display a blank page when there are multiple jobs.
-      while (item.updating) {
-        await sleep(100)
-      }
     }
 
     return true
--- a/lib/openers/pdf-view-opener.js
+++ b/lib/openers/pdf-view-opener.js
@@ -3,24 +3,34 @@
 import Opener from '../opener'
 import { isPdfFile } from '../werkzeug'
 
-function forwardSync (pdfView, texPath, lineNumber) {
-  if (pdfView != null && pdfView.forwardSync != null) {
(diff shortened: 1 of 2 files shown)


lib/openers/pdf-view-opener.js
```diff
@@ -3,24 +3,34 @@
...
+    }
+
+    const item = await atom.workspace.open(filePath, options)
+    const pane = atom.workspace.paneForItem(item)
+    if (pane && item && item.forwardSync) {
+      item.forwardSync(texPath, lineNumber)
+      // If we don't wait for pdf-view to do it's first update then it will just
+      // display a blank page when there are multiple jobs.
+      while (item.updating) {
+        await sleep(100)
```
thomasjo: While I really understand the attractiveness of this solution, it is very fragile and will not work as expected on e.g. slow computers, etc. Having this as a temporary hack is OK, but let's document it.

The proper, long term solution is to open a PR against pdf-view, exposing a callback (or promise) that resolves the problem we're encountering.

yitzchak: I agree. It's just a temporary hack. I'll submit a PR to `pdf-view` and leave this PR open until there is an obvious resolution. They were pretty fast with the last PR I submitted. It will probably have to be done with a callback or `EventEmitter` though. There are some time delays used in `pdf-view`.

yitzchak: Looks like the issue is related to `Fit to width on open` setting in `pdf-view`. When that option is off everything works. When it is on then the resize fails when the tab is not visible. I'll remove the hack and investigate the fix for `pdf-view` or submit a bug report if I can't fix it.

Also, when not using DiCy activating the center pane before splitting isn't working because only the first job gets file name and line number for sync. It appears that this is because `showResult` is calling `getEditorDetails` on a `pdf-view`. I'll fix that in a separate PR since the fix will involve changes to `Composer`.

package.json
```diff
@@ -255,37 +255,49 @@
       "default": "automatic",
       "order": 17
     },
+    "openerSplit": {
```
yitzchak: @thomasjo Do you have preference regarding naming? 

- `openerSplit`
- `openerSplitDirection`
- something else?

thomasjo: You know me, I like explicit and verbose names; my vote is for `openerSplitDirection`. Can't really think of a better alternative.

thomasjo: On second though, since this is only support by pdf-view, perhaps something like `pdfViewSplitDirection` (or `pdfViewSplit`). Alternatively we simply add a description detailing that this option is currently only supported by the pdf-view opener? _The Principle of Least Surprise™_

yitzchak: Ok, how about `pdfViewSplitDirection` and we can change it to `openerSplitDirection` if we actually add other viewers and we want to have a single unified setting?

thomasjo: Sounds good!

PR #401, opened 2017-08-15
Extends requirements to properly function

pabloguerrero:
This is required to make working with Latex documents much more usable. Also, the keybindings didn't work for me when the language-latex package was not installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,10 @@ selecting the `Use DiCy` option in the settings page. [DiCy][] will be installed
 automatically and so no further action is required for either TeX Live or
 MiKTeX.
 
+### Code Highlighting
+In order to properly work in Atom with this package, you'll need a code
+highlighting extension, namely [`language-latex`][language-latex-pkg].
+
 ## Usage
 The `latex:build` command can be invoked from the LaTex menu or by pressing the
 default keybind <kbd>ctrl</kbd>-<kbd>alt</kbd>-<kbd>b</kbd> while in a LaTex or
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,6 +53,10 @@ selecting the `Use DiCy` option in the settings page. [DiCy][] will be installed
 automatically and so no further action is required for either TeX Live or
 MiKTeX.
 
+### Code Highlighting
```
thomasjo: Rename this to "Syntax Highlighting" to better match the terminology used within Atom.

README.md
```diff
@@ -53,6 +53,10 @@ selecting the `Use DiCy` option in the settings page. [DiCy][] will be installed
...
+In order to properly work in Atom with this package, you'll need a code
+highlighting extension, namely [`language-latex`][language-latex-pkg].
+
```
thomasjo: I propose that this paragraph be changed to the following:
``` md
In order for this package to behave as expected, your Atom environment must
contain a package that provides a LaTeX grammar. We suggest
[`language-latex`][language-latex-pkg], but other valid options might exist.
``` 
What do you think? /cc @yitzchak

thomasjo: Additionally, upon discussing this with @yitzchak, we think it makes sense to recommend using the language-knitr package as well. Maybe an addition to the above, along the lines of:
``` md
Additionally, you might want to consider installing
[`language-knitr`][language-knitr] if you want to take advantage of the
provided knitr support.
``` 

@yitzchak Should we also mention language-r perhaps? Not strictly required, but highly recommended in combination with language-knitr.

yitzchak: I would definitely add the `language-r` mention. I don't think it we have to explain that `language-knitr` imports `language-r` (and `language-latex`) since knitr users will understand that R is required.

```md
Additionally, you might want to consider installing
[`language-knitr`][language-knitr] and [`language-r`][language-r]
if you want to take advantage of the provided knitr support.
```

yitzchak:
@pabloguerrero We have added other build paths which will need to be documented along with additional syntax package requirements. To keep things simple I have updated this PR with the prose changes in the review requests so we can merge it.

Thanks again for helping out. We really appreciate it!

Issue #342, opened 2017-02-02
Only build on saving .tex files


joernschellhaas:
As an enhancement of #150 it would be great if build on save was limited to some file types (e. g. .tex)

Additionally/alternatively, there could be a `%!TEX buildonsave = true` directive or similar.

yitzchak:
Thank you for the feature request! We'll look into it.

yitzchak:
This enhancement will be in the next version (based on file extension).

PR #408, opened 2017-09-05
Update DiCy

yitzchak:
DiCy v0.9.1 is more strict and adds option value validation. Update `Composer` as needed. Also update file extension filters to permit Pweave and literate Agda.

### DIFF
--- a/lib/composer.js
+++ b/lib/composer.js
@@ -322,12 +322,12 @@ export default class Composer extends Disposable {
     latex.status.setIdle()
   }
 
-  async build (shouldRebuild = false, silent = false) {
+  async build (shouldRebuild = false, enableLogging = true) {
     await this.kill()
 
     const { editor, filePath, lineNumber } = getEditorDetails()
 
-    if (!this.isValidSourceFile(filePath, silent)) {
+    if (!this.isValidSourceFile(filePath, enableLogging)) {
       return false
     }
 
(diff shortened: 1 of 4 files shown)


yitzchak:
@thomasjo Can you take a look at this one?

yitzchak:
Also resolves #342

lib/composer.js
```diff
@@ -318,13 +322,12 @@ export default class Composer extends Disposable {
     latex.status.setIdle()
   }
 
-  async build (shouldRebuild) {
+  async build (shouldRebuild = false, silent = false) {
```
thomasjo: I think `quiet` is more frequently used, but what about making this more verbose and understandable; `suppressLogging`. Alternatively, flip the logic and name it `verbose` or some such.

lib/composer.js
```diff
@@ -333,7 +336,7 @@ export default class Composer extends Disposable {
     }
 
     if (this.shouldUseDiCy()) {
-      return this.runDiCy(['load', 'build', 'log', 'save'], { ignoreCache: shouldRebuild },
+      return this.runDiCy(['load', 'build', 'log', 'save'], { ignoreCache: !!shouldRebuild },
```
thomasjo: This seems unnecessary given the argument has a default value of `false`.

yitzchak: Good point.

lib/composer.js
yitzchak: What do think about `enableLogging`?

thomasjo: I like it :1st_place_medal:

PR #416, opened 2017-09-16
Fix pdf-view focus

### LINKED ISSUE #413: Open with pdf-view doesn't open in background
I have "Open Result in Background" option ticked and I've selected "Opener" to be `pdf-view` and "Pdf View Split Direction" to `right`. I also have "Build in save" and "Open Result after Successful Build" ticked.

It correctly builds the PDF on save and opens it in the right panel, but the focus now is on the PDF and not on the editor pane. Which means I have to manually focus the edit pane every time I press `ctrl+s`.

I believe "Open Result in Background" means that this should not happen and after opening the PDF in right pane, the focus should be retained at the editor pane.

`latex` version installed is `0.46.0`

```
$ atom --version
Atom    : 1.19.3
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

yitzchak:
Respect `openInBackground` setting by saving and re-activating previous pane. Resolves #413

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased][] / YYYY-MM-DD
+### Added
+- Support for opening result in background with `pdf-view`. Resolves
+  [413](https://github.com/thomasjo/atom-latex/pull/413).  
+    ~ [#416](https://github.com/thomasjo/atom-latex/pull/416)
+    / [@nisargjhaveri](https://github.com/nisargjhaveri)
+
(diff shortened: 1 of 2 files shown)


thomasjo:
Would be awesome to have some tests for this both to ensure we aren't breaking anything and also to guard against future regressions, but I'm not sure it's worth the effort right now. Would likely involve a fair bit of orchestration and pane wizardry :sparkles:

yitzchak:
I agree. UI tests would be awesome but seem like a lot of work.

I planning updating the change with credit to the issue creator since he did all the work, FYI.

thomasjo:
> I am planning on updating the change log with credit to the issue creator since he did all the work, FYI.

Yeah, was just about to suggest that :smiley:

PR #420, opened 2017-09-23
Fix marker reset issue

### LINKED ISSUE #419: Warning/Error bars persist after successful compilation
Atom 1.20.1, on OSX.
Once I create an error, I can fix it and the visual marker adjacent to line numbers remains after I recompile. This can be annoying because it still makes me think I have errors at first glance, and it's harder to tell if I've fixed it, especially if I have multiple errors and only fixed some of them.

screen shot 2017-09-22 at 10 13 52 am

See line 72 on the screenshot, and notice there are 0 errors in at the bottom compilation status.

The same holds for warning messages.

I can remove the marker (after noticing that it persists) by changing the line(s) where it occurs.

yitzchak:
The reset flag is getting lost because Atom's Emitter only handles one event value. Fix this by wrapping the event data with a single object and destructure in the event handlers. Fixes #419.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
     ~ [#416](https://github.com/thomasjo/atom-latex/pull/416)
     / [@nisargjhaveri](https://github.com/nisargjhaveri)
 
+### Fixed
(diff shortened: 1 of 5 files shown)


yitzchak:
@thomasjo Added more tests for `Logger` and `MarkerManager`.

Issue #430, opened 2017-10-19
Log warning line numbers for included files


hendrikbl:
I'm using `\include` a lot to structure my project. When there is a warning in an included file the log only shows the position where the file is included in the root file. Is there any way of logging the actual position of the warning so i dont have to look through the entire file I included?

yitzchak:
Thanks for reporting the issue!

If a line number is reported in a warning message then it should be included in the reference from the log panel. That said, it doesn't always work because log messages produced by TeX are incredibly varied.

If you a simple document that demonstrates the issue that help enormously.

hendrikbl:
Thanks for the quick response! 

Here's an example:

### main.tex
This is the file with the include in it
```latex
\begin{document}
\frontmatter
\tableofcontents
\mainmatter
\include{include}
\end{document}
```

### include.tex
This is the file with an underfull hbox
```latex
\begin{titlepage}
	\dotfill\\
	Prof. XYZ\\
	Department of XYZ\\
	University of whatever, \\
	City, Country\\

	Date: ............\\
\end{titlepage}
```

But the log says the warning is in `main.tex (2-7)` 
![warning message](https://user-images.githubusercontent.com/9558391/31768353-52e15f0e-b4cf-11e7-9f27-4c3e5ef7c6b3.PNG)

I'm not sure if thats something you can fix though or if it's on TeX related.

yitzchak:
Thanks for sample document. I have reproduced the issue.

It looks like TeX is putting a file reference on an earlier line (in the log) before the group of box messages. Not sure if it consistent in that yet. Fixing it will require enhancing the log parser so it will take a bit work to resolve.

hendrikbl:
Okay, I already guessed it isn't something that can be fixed easily...

Thanks for looking into it!

yitzchak:
This enhancement will be be in the next version. We just merged the code for the `latexmk` based builder and support for DiCy is in the works.

yitzchak:
Added in both builders and released in v0.49.0

PR #432, opened 2017-10-19
Parse input source markers in log

yitzchak:
This is an experimental PR that attempts to (somewhat naively) parse the markers in the log from `\input` and `\include` so as to more accurately determine the source file for warning messages.

The motivating issue is #430.

### DIFF
--- a/lib/parsers/log-parser.js
+++ b/lib/parsers/log-parser.js
@@ -56,9 +56,11 @@ export default class LogParser extends Parser {
     }
     const sourcePaths = [this.texFilePath]
 
-    // Ignore the first line because it has some confusing patterns
-    const lines = this.getLines().slice(1)
+    const lines = this.getLines()
     lines.forEach((line, index) => {
+      // Ignore the first line because it has some confusing patterns
+      if (index === 0) return
+
       // Simplest Thing That Works™ and KISS®
       const logRange = [[index, 0], [index, line.length]]
       let match = line.match(OUTPUT_PATTERN)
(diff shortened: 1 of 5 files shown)


yitzchak:
@thomasjo Take a look at this please. Thoughts?

yitzchak:
So far I have found only one LaTeX document that has some issues. The file references in the document all get parsed properly so it doesn't really represent a failure in the algorithm, but ignoring the first line of the log file and protecting the first item in the stack with `if (sourcePaths.length > 1)` appear to be required.

Unfortunately it is a large document at about 500 pages, so it will take a little bit of work to go through the log file and see if the issue can be distilled to a test case. I suspect that there is an opening parenthesis in a package message and a closing parenthesis on a following line that is not getting captured.

FYI, compiling the document requires DiCy since it has a lot of external dependencies. 

Also, FYI the DiCy version of this PR [yitzchak/dicy#103].

yitzchak:
Looks the issue was a DiCy only issue. Basically, DiCy parses continued package messages like the following.

```
Package rerunfilecheck Info: File `book.out' has not changed.
(rerunfilecheck)             Checksum: 20C53CC462287330CC11BBD2C02A09A2;6814.
```

To avoid intepreting `(./ppivot.aux) (./network.aux) (./vs1.aux) (./vs2.aux) (./vs3.aux)` as a continued message I just had to strengthen the pattern a bit. `LogParser` is not currently  parsing these continuations so it is a non-issue when using `latexmk`.

PR #439, opened 2017-11-07
Migrate specs to use async helpers

thomasjo:
I shamelessly procured the [`async-spec-helpers.js`](https://raw.githubusercontent.com/atom/atom/master/spec/async-spec-helpers.js) file from Atom, and migrated all of our specs to leverage those async helpers.

This PR represents a small step towards eventually upgrading Jasmine to a modern release, or perhaps migrating to Mocha + Sinon and what not.

All specs (except SageTeX and Asymptote) are green on my Ubuntu 17.04 system, using TeX Live 2017.

### DIFF
--- a/script/install-knitr
+++ b/script/install-knitr
@@ -8,6 +8,10 @@ shopt -s nocasematch
 
 ensure_r_installed() {
   if [[ "${APPVEYOR:-}" == true ]]; then
+    # TODO: Remove when 'R.Project' package adds it as a dependency.
+    echo "Installing chocolatey-core.extension..."
+    choco install chocolatey-core.extension
+
     echo "Installing R..."
     choco install R.Project --allow-empty-checksums --ia="/DIR=\"${R_HOME}\""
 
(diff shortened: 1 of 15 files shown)


thomasjo:
AppVeyor failure is due to R installation through Chocolatey being broken. Looks like it might be a breaking change in Chocolatey, or something along those lines. Either way, unrelated to this PR. Assuming it passes on Travis, I will consider that as :green_heart:

yitzchak:
I think that the AppVeyor failure is due to the R package is missing a dependency on `chocolatey-core.extension`. GhostScript includes that dependency so installing GhostScript before knitr should fix this issue. DiCy installs in the reverse order and it seems to work.

We could also add `choco install chocolatey-core.extension` before the R installation. Either method should work to give us :green_heart: if you don't want to wait for the package to be fixed.

https://chocolatey.org/docs/helpers-get-package-parameters

yitzchak:
SageTeX and Asymptote are green on my Arch box.

thomasjo:
@yitzchak Thanks for the ProTip™ on how to resolve the AppVeyor build failures :bowing_man:

yitzchak:
@thomasjo No problem at all. The specs are much more readable with your async updates, BTW.

PR #442, opened 2017-11-17
Added qpdfview with synctex 

skidpa:
Hi i created a opener for qpdfview with synctex and appendend the package.json file with what i think is the appropriate information. It's really just a modified zathura-opener

this is my first time contributing so i might have made a fool of my self though.

### DIFF
--- a/lib/openers/qpdfview-opener.js
+++ b/lib/openers/qpdfview-opener.js
@@ -9,7 +9,7 @@ export default class QpdfviewOpener extends Opener {
     const qpdfviewPath = atom.config.get('latex.qpdfviewPath')
     const args = [
       `--unique`,
-      `${filePath}#src:${texPath}:${lineNumber}:0`
+      `"${filePath}"#src:"${texPath}":${lineNumber}:0`
     ]
     const command = `"${qpdfviewPath}" ${args.join(' ')}`
     await latex.process.executeChildProcess(command, { showError: true })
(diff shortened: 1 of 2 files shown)


lib/openers/qpdfview-opener.js
```diff
@@ -0,0 +1,27 @@
...
+import fs from 'fs-plus'
+import Opener from '../opener'
+import { isPdfFile, isPsFile } from '../werkzeug'
+
+export default class QpdfviewOpener extends Opener {
+  async open (filePath, texPath, lineNumber) {
+    const qpdfviewPath = atom.config.get('latex.qpdfviewPath')
+    const args = [
+      `--unique`,
+      `${filePath}#src:${texPath}:${lineNumber}:0`
```
yitzchak: You need to quote the file argument in case there are spaces in the path.

yitzchak:
Thank you for submitting the PR. We really appreciate any and all code submissions!

So far I just see the issue regarding quoting the file path argument, which I've added a line comment for that.

@thomasjo Can you take a look when you get a chance?

skidpa:
Your are absolutely right. I shamelessly overlooked that part, it should be
`"${filePath}"#src:"${texPath}":${lineNumber}:0`
or 
`"${filePath}#src:${texPath}":${lineNumber}:0`
both seems to work fine on my system with spases in the paths's and without i guess its mostly a matter of which way you preferr.
I've fixed it and pushed the first option to my repo anything else i should do?

thomasjo:
Thanks for adding this! 🥇

PR #447, opened 2017-12-09
Add tikz extension

### LINKED ISSUE #444: Magic TeX comments are not interpreted in TikZ files
Magic TeX comments are not interpreted in TikZ files. If I have a TikZ file and set the TeX comment to `%	!TEX root = main.tex` in order to have the main file being compiled, after starting the compilation I get the output `The file does not appear to be a valid source file.`

If I rename the Tikz file to `*.tex`, the TeX comment works. This used to be different in previous versions of the package, but I don't know when exactly this broke. Was this change by accident or intentional?

Please also see the minimal (not) working example. If you rename tikz to tex, the main file will be compiled when starting compilation on the external file.

MWE:
file `main.tex`:
```
\documentclass[]{article}

\usepackage{pgfplots}

\begin{document}
Magic TeX comments are not interpreted in TikZ files.

\input{external.tikz}

\end{document}
```

file `external.tikz`:
```
%	!TEX root = main.tex
\begin{tikzpicture}

\draw [-,style=-, line width=0.5mm] (0,0) -- ++(8,0);

\end{tikzpicture}%
```

yitzchak:
Fixes #444

### DIFF
--- a/lib/werkzeug.js
+++ b/lib/werkzeug.js
@@ -37,7 +37,7 @@ export default {
   },
 
   isSourceFile (filePath) {
-    return filePath && !!filePath.match(/\.(?:tex|lhs|lagda|[prs]nw)$/i)
+    return filePath && !!filePath.match(/\.(?:tex|tikz|lhs|lagda|[prs]nw)$/i)
   },
 
   isTexFile (filePath) {


raimund-schluessler:
I can confirm that this fixes the issue. Thanks a lot for having a look. Sorry I saw this so late.

PR #448, opened 2017-12-09
Fix Rscript in AppVeyor

yitzchak:


thomasjo:
Any idea why this is suddenly breaking? This has worked for a long time. By splitting this across files, we're spreading crucial, inter-dependent steps across multiple files.

Adding the hack to `.appveyor.yml` is fine if it's truly the only solution, but removing the `PATH` step entirely from `install-knitr` is not OK. The idea of the CI scripts is that they should work, with minimal effort, on any build system. I tested these on a Windows VM back in the day by simply setting a few environment variables to "mimic" AppVeyor. If I somehow have to "execute" the YAML file (or run more manual steps), things not that easy.

I don't like changing things the easy way simply to get stuff working. That quickly becomes a house of cards. Then again, if we have no other viable alternative, keep the addition, but remove the deletion.

.appveyor.yml
```diff
@@ -28,7 +28,7 @@ cache:
   - '%R_LIBS_USER%'
 
 build_script:
-  - set PATH=C:\Program Files\Git\mingw64\bin;%PATH%
+  - set PATH=C:\Program Files\Git\mingw64\bin;%R_HOME%\bin;%PATH%
```
thomasjo: This is OK if there's no other viable alternative, but this is a "hack" and we should probably even have it on a separate line and mark it as a (temporary) hack via a comment.

script/install-knitr
```diff
@@ -14,8 +14,6 @@ ensure_r_installed() {
 
     echo "Installing R..."
     choco install R.Project --allow-empty-checksums --ia="/DIR=\"${R_HOME}\""
-
-    export PATH="${R_HOME}/bin:${PATH}"
```
thomasjo: Keep this.

yitzchak:
I don't know why this happening. The same set of install steps in DICy seem to work. The only difference was the path setting which why I changed it. The hack doesn't seem to work so we are back to square one.

yitzchak:
It looks like this is a bug in R. I got a green board by removing the spaces from all calls to `Rscript`. I believe that bug was introduced in [r73703](https://svn.r-project.org/R/trunk/src/gnuwin32/front-ends/Rfe.c?p=r73703) (v3.4.3) and *may* be resolved by [r73849](https://svn.r-project.org/R/trunk/src/gnuwin32/front-ends/Rfe.c?p=r73849) which has not landed in a release yet.

What would you like to do, get rid of the spaces in `Rscript` calls or wait for a fix?

yitzchak:
@thomasjo Can you take a look at this when you get a chance?

yitzchak:
Bug fixed upstream.

Issue #406, opened 2017-08-30
Log report (on bottom bar) appears on non-latex files


lzkelley:
I'm not sure if this is a bug or feature request, but currently the log report (build status, errors, warnings, info --- see pic) appears even when a non-latex file is open.  It would be great if the latex report did not appear in that case.

![screen shot 2017-08-30 at 11 09 31 am](https://user-images.githubusercontent.com/3839265/29879728-ee100fa8-8d73-11e7-8206-a49d630207fb.png)

yitzchak:
Thanks for feature request! We'll look into it.

eboracus:
I just came here to report this same issue. It seems that it is currently necessary to disable/enable the entire package to have it only appear at appropriate times. 😞

Dissapointed to see the stale bot has been by for a visit, could it be requested to make this 'fresh' again? (I almost submitted a duplicate issue through missing this one).

aubreyz:
I'd +1 this issue. The appearance of the bottom Latex toolbar in every context in atom is the only real beef I have with this package which does a great and uncomplicated job.

PR #457, opened 2018-03-08
Only show the status bar label for latex files

pwxn:
Should help with issue #406:

Whenever the active editor is changed, this decides to attach/detach label from the status bar based on the filepath of the editor. Specifically, if the path extension contains "tex".

(This is my first pull request to an atom package, feedback is very welcome, in case I've broken the best practices)

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -45,7 +45,9 @@ export default {
     }
 
     atom.workspace.onDidChangeActivePaneItem((editor: AtomCore.IEditor) => {
-      this.checkStatusLabel(editor.getPath())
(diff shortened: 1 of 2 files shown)


pwxn:
I fixed the check to use the isSourceFile() function instead of naively matching "tex" in the filepath.

pwxn:
I've looked at the logs on Travis, but I don't think the failures are related to this PR.

thomasjo:
Thanks for the contribution! I'll take a look at this during the next few days (most likely the weekend).

pwxn:
Closing old

PR #459, opened 2018-03-14
Fix js-standard lint issues

yitzchak:


lib/builder.js
```diff
@@ -1,4 +1,5 @@
 /** @babel */
+/* eslint-disable no-multi-spaces */
```
thomasjo: What's the rationale behind this? Has something changed in JavaScript Standard Style recently?

script/install-knitr
```diff
@@ -19,7 +19,6 @@ ensure_r_installed() {
   elif [[ "${TRAVIS_OS_NAME:-}" == "osx" ]]; then
     echo "Installing R..."
     brew update
-    brew tap homebrew/science
```
thomasjo: Why is this being removed in this PR? Is this causing errors during CI?

lib/composer.js
```diff
@@ -216,38 +216,39 @@ export default class Composer extends Disposable {
...
+        latex.log.showMessage(message)
+      })
+      .on('command', event => {
+        latex.log.info(`[${event.rule}] Executing \`${event.command}\``)
+      })
+      .on('fileDeleted', event => {
+        if (!event.virtual) {
+          latex.log.info(`Deleting \`${event.file}\``)
+        }
+      })
```
thomasjo: I'd prefer it if all of these `.on` calls were not chained like this. Right now the code is pretty difficult to read. Each `.on` call should a separate `dicy.on` call, perhaps with a blank line separating each of them.

script/install-knitr
yitzchak: Yes, homebrew/science is deprecated. R appears to be available without the tap.

lib/builder.js
yitzchak: The linter is complaining about multiple spaces before comments. Without the override the columns won't line up, unless you want to override it for the whole project?

lib/composer.js
yitzchak: Sounds good.

lib/builder.js
thomasjo: Not worth keeping the alignment for this file. For the parser files that contain aligned EOL comments, we should disable the rule only for those blocks of regex patterns, and re-enable immediately after them.

lib/composer.js
```diff
@@ -229,9 +229,7 @@ export default class Composer extends Disposable {
         if (event.source.range) {
           message.range = [[event.source.range.start - 1, 0], [event.source.range.end - 1, Number.MAX_SAFE_INTEGER]]
         }
-      }
-
-      if (event.log) {
+      }      if (event.log) {
```
thomasjo: This is why disabling the rule carte blanche is not an ideal solution. Looks like these extra spaces should be preceded with a newline.

package.json
```diff
@@ -22,17 +22,17 @@
     "etch": "0.12.6",
     "fs-plus": "^3.0.0",
     "glob": "^7.1.1",
-    "js-yaml": "^3.8.4",
+    "js-yaml": "^3.11.0",
```
thomasjo: In general I'd prefer it if we don't mix in random package upgrades in PRs which are supposed to resolve something entirely unrelated, but let's keep it like this for this PR.

Issue #462, opened 2018-03-21
Default clean patterns cleans too many files


Doekeb:
### Description

The "clean" command cleans all files in the current directory which begin with the name of the current file and have an extension listed in the clean patterns setting. The MWE and the discussion under "Additional Information" should make it clear what is happening and why this is a bad problem.

### Steps to Reproduce

0. Install the latex package in atom, keep default options.
1. Navigate to (or create) an empty directory.
2. Create three files, save them as "test.tex", "testing.tex", and "other.tex".
3. Write enough LaTeX in each file to build a pdf for each of these files. For instance, you could paste the following:
```
\documentclass{article}

\begin{document}
     Lorem ipsum dolor.
\end{document}
```
4. Use the "Latex: Build" command on all three files (keyboard shortcut `ctrl-alt-b`).
5. Place the cursor in the file "test.tex".
6. Use the "Latex: Clean" command (keyboard shortcut `ctrl-alt-c`). Make sure to only use this command once, and when the cursor is in the file "test.tex".

**Expected behavior:**
The only files which are deleted are those which were created from using the "Latex: Build" command on the file "test.tex". In this MWE, the complete list of deleted files should be:

test.aux
test.fdb_latexmk
test.fls
test.log
test.pdf
test.synctex.gz

*All other files should remain intact and undeleted*

**Actual behavior:**
All files which start with "test" and have extensions in the clean patterns list are deleted. These include the following:

testing.aux
testing.fdb_latexmk
testing.fls
testing.log
testing.pdf
testing.synctex.gz

**Reproduces how often:**
100%

### Versions

Atom : 1.25.0
latex: 0.49.0
macOS: 10.13.2

### Additional Information

Here are two example scenarios which show why this behavior is undesirable.

1. You write a math paper and save it as "My great paper.tex". It relies on an important graphic which you've spent hours perfecting, and is saved as "My great paper graphic.pdf" in the same directory. The file "My great paper.tex" isn't compiling and you can't figure out why, so you decide to clean the auxiliary files. To your great surprise, the all important "My great paper graphic.pdf" has disappeared!

[...]

PR #463, opened 2018-03-28
Fix incorrect glob pattern used to populate list of generated files.

thomasjo:
Bug discovered by @Doekab in #462. This bug has been around for a very long time, but somehow nobody has encountered it before now. The basis gist of the bug is as follows;

1. Given that `file.tex` is a valid LaTeX document that compiles, and that the build process produces some auxiliary files, of which one is `file.pdf`.
2. Rename `file.pdf` to `file x.pdf`.
3. When executing the `latex:clean` command, the renamed `file x.pdf` is incorrectly cleaned due to the overzealous glob pattern.

The proposed fix is simply replacing the very greedy glob pattern with a slightly less greedy pattern that only does wildcard matching on file extensions. Note that strictly speaking, the proposed fix is probably still susceptible to issues with files named e.g. `file.some.thing.log`, but that will have to be verified and fixed at some later stage.

### DIFF
--- a/lib/composer.js
+++ b/lib/composer.js
@@ -509,7 +509,7 @@ export default class Composer extends Disposable {
       builder.parseLogAndFdbFiles(jobState)
     }
 
-    const pattern = path.resolve(dir, jobState.getOutputDirectory(), `${jobState.getJobName() || name}*`)
+    const pattern = path.resolve(dir, jobState.getOutputDirectory(), `${jobState.getJobName() || name}.*`)
     const files = new Set(glob.sync(pattern))
     const fdb = jobState.getFileDatabase()
 
--- a/spec/composer-spec.js
+++ b/spec/composer-spec.js
@@ -292,6 +292,18 @@ describe('Composer', () => {
       expect(fs.removeSync).toHaveBeenCalledWith(path.join(fixturesPath, '_minted-wibble'))
     })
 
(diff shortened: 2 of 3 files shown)


thomasjo:
As a side note, while working on this I noticed that our clean-related tests are quite lacking, and due to the use of the mocked `getGeneratedFileList` implementation is likely hiding many other bugs. We should strongly reconsider how and when we're mocking crucial infrastructure. At the very least, if we want to keep the current mock, we should add several positive and negative test that only target `Composer.getGeneratedFileList` to ensure we're fully covering all scenarios.

lib/composer.js
```diff
@@ -509,7 +509,7 @@ export default class Composer extends Disposable {
       builder.parseLogAndFdbFiles(jobState)
     }
 
-    const pattern = path.resolve(dir, jobState.getOutputDirectory(), `${jobState.getJobName() || name}*`)
+    const pattern = path.resolve(dir, jobState.getOutputDirectory(), `${jobState.getJobName() || name}.*`)
```
yitzchak: There are a few automatically generated files that this pattern will miss. Specifically, BibTeX generates `{jobName}-blx.bib` and RevTeX generates `{jobName}Notes.bib`. Not sure if these are worth including or just relying on the existence of a fdb file instead.

yitzchak (commented):
Please see comment regarding file name pattern.

PR #496, opened 2018-11-11
Show Name of Undefined Control Sequence (macro)

Strauman:
When LaTeX complains about undefined control sequence, this plugin does
not show the name of said macro, which is what this commit attempts to
fix.

### DIFF
--- a/lib/parsers/log-parser.js
+++ b/lib/parsers/log-parser.js
@@ -21,7 +21,7 @@ const ERROR_PATTERN = new RegExp('' +
 // Pattern for obtaining the last macro in a line
 // Meant for catching the undefined control sequence
 const LAST_MACRO_PATTERN = new RegExp('' +
-  '(\\\\[a-zA-Z@]+)$' // Macro name
+  '(\\\\[^\\\\]+)$' // Macro name
 )
 
 // Pattern for overfull/underfull boxes
--- a/lib/parsers/log-parser.js
+++ b/lib/parsers/log-parser.js
@@ -85,27 +85,28 @@ export default class LogParser extends Parser {
 
       match = line.match(ERROR_PATTERN)
       if (match) {
-        let outMessageText
+        // Will contain the entire message that is going to be shown to the user
+        let outputMessage
         const lineNumber = match[2] ? parseInt(match[2], 10) : undefined
         // Check if the error is "Undefined control sequence"
-        let errMessageText = match[4]
-        if (errMessageText === UNDEF_CONTROL_SEQ_TEXT) {
+        const errorMessage = match[4]
+        if (errorMessage === UNDEF_CONTROL_SEQ_TEXT) {
           // The actual macro is the last thing
           // in the sentence of the next line
-          let nextLine = lines[index + 1]
-          let macroMatch = nextLine.match(LAST_MACRO_PATTERN)
+          const nextLine = lines[index + 1]
+          const macroMatch = nextLine.match(LAST_MACRO_PATTERN)
           // Check if there is a match
           if (macroMatch) {
-            outMessageText = errMessageText + ': ' + macroMatch[1]
+            outputMessage = `${errorMessage}: ${macroMatch[1]}`
           } else {
-            outMessageText = 'Unknown undefined controlsequence'
+            outputMessage = 'Unknown undefined controlsequence'
           }
         } else {
-          outMessageText = (match[3] && match[3] !== 'LaTeX') ? match[3] + ': ' + match[4] : match[4]
+          outputMessage = (match[3] && match[3] !== 'LaTeX') ? match[3] + ': ' + match[4] : match[4]
         }
         result.messages.push({
           type: 'error',
-          text: outMessageText,
+          text: outputMessage,
           filePath: match[1] ? path.resolve(this.projectPath, match[1]) : sourcePaths[0],
           range: lineNumber ? [[lineNumber - 1, 0], [lineNumber - 1, Number.MAX_SAFE_INTEGER]] : undefined,
           logPath: this.filePath,
--- a/lib/parsers/log-parser.js
+++ b/lib/parsers/log-parser.js
@@ -21,7 +21,7 @@ const ERROR_PATTERN = new RegExp('' +
 // Pattern for obtaining the last macro in a line
 // Meant for catching the undefined control sequence
 const LAST_MACRO_PATTERN = new RegExp('' +
-  '[^\\\\]*(\\\\[a-zA-Z@]+)$' // Macro name
+  '(\\\\[a-zA-Z@]+)$' // Macro name
 )
 
 // Pattern for overfull/underfull boxes
--- a/lib/parsers/log-parser.js
+++ b/lib/parsers/log-parser.js
@@ -18,6 +18,12 @@ const ERROR_PATTERN = new RegExp('' +
   '(.+?)\\.?$'           // Message text, the ending period is optional for MiKTeX
 )
 
+// Pattern for obtaining the last macro in a line
+// Meant for catching the undefined control sequence
+const LAST_MACRO_PATTERN = new RegExp('' +
+  '[^\\\\]*(\\\\[a-zA-Z@]+)$' // Macro name
+)
+
 // Pattern for overfull/underfull boxes
 const BOX_PATTERN = new RegExp('' +
   '^((?:Over|Under)full \\\\[vh]box \\([^)]*\\))' + // Message text
@@ -31,6 +37,9 @@ const WARNING_INFO_PATTERN = new RegExp('' +
   '(.*?)' +                                           // Message text
   '(?: on input line (\\d+))?\\.$'                    // Line number
 )
+
+// LaTeX log text when macro is undefined
+const UNDEF_CONTROL_SEQ_TEXT = 'Undefined control sequence'
 
 /* eslint-enable no-multi-spaces */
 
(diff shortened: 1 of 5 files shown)


Strauman:
Simple LaTeX test case:

```
\documentclass{minimal}
\def\once{Once}
\begin{document}
  \once upon a time there was an \undefined control sequence.
\end{document}
```

Expected log output:
screen shot 2018-11-11 at 11 52 27

lib/parsers/log-parser.js
```diff
@@ -18,6 +18,12 @@ const ERROR_PATTERN = new RegExp('' +
   '(.+?)\\.?$'           // Message text, the ending period is optional for MiKTeX
 )
 
+// Pattern for obtaining the last macro in a line
+// Meant for catching the undefined control sequence
+const LAST_MACRO_PATTERN = new RegExp('' +
+  '[^\\\\]*(\\\\[a-zA-Z@]+)$' // Macro name
```
Strauman: I'm not sure as to what would constitute a legal macro name. 

One could argue that it _might_ be better to just look at the last
space and go from there.

An issue with the definition here is that I din't include underscores `_`, 
which will be very widely used in macro names come LaTeX3.

```suggestion
  '[^\\\\]*(\\\\[a-zA-Z@_]+)$' // Macro name
```

lib/parsers/log-parser.js
```diff
@@ -18,6 +18,12 @@ const ERROR_PATTERN = new RegExp('' +
   '(.+?)\\.?$'           // Message text, the ending period is optional for MiKTeX
 )
 
+// Pattern for obtaining the last macro in a line
+// Meant for catching the undefined control sequence
+const LAST_MACRO_PATTERN = new RegExp('' +
+  '(\\\\[a-zA-Z@]+)$' // Macro name
```
Strauman: I'm not confident as to what ideally should constitute a legal macro name in
this setting.

One of the biggest gotchas here is that in LaTeX, even space can be made
a letter-character (thus allowed in macro names). However, that would be
weirdly inconvenient. The other is that any character can be made catcode 0,
that is, the escape character (by default `\`).

I would argue that it's be better to just look at the last space and
go from there, as it is rather unlikely that one would redefine the space
as such:

```suggestion
  '([^ ]+)$' // Macro name
```

An issue with the definition here is that I din't include underscores `_`, 
which will be very widely used in macro names come LaTeX3.

```suggestion
  '(\\\\[a-zA-Z@_]+)$' // Macro name
```

Strauman: Actually, I think the best way would be to assume that `\` is the escapechar and

```suggestion
  '(\\\\[^\\\\]+)$' // Macro name
```

Strauman:
Probably a test should be made to see if there is any matches!

lib/parsers/log-parser.js
```diff
@@ -18,6 +18,12 @@ const ERROR_PATTERN = new RegExp('' +
   '(.+?)\\.?$'           // Message text, the ending period is optional for MiKTeX
 )
 
+// Pattern for obtaining the last macro in a line
+// Meant for catching the undefined control sequence
+const LAST_MACRO_PATTERN = new RegExp('' +
+  '(\\\\[^\\\\]+)$' // Macro name
```
Strauman: I'm not highly confident as to what ideally should a legal macro name in
this setting.

The problem is that the escape character, `\` by default, can be redefined, as well
as having anything being active letters (and thus treated as control sequences)

E.g. doing
```
\catcode`\O=13\relax O
```
Would make the letter O an unknown control sequence, or

```
\catcode`J=0\relaxJUndefined
```
Would make `JUnedfined` behave identically to `\Undefined` and this atom plugin
will report weird information.

You can't just look at the last space and go from there. If you consider
```
{}\some@undef@macro
```
Then the `{}` would be a part of the macro name.

However, I do argue that these cases are extraordinary circumstances,
and isn't really that necessary to implement.

lib/parsers/log-parser.js
```diff
@@ -76,10 +85,27 @@ export default class LogParser extends Parser {
 
       match = line.match(ERROR_PATTERN)
       if (match) {
+        let outMessageText
         const lineNumber = match[2] ? parseInt(match[2], 10) : undefined
+        // Check if the error is "Undefined control sequence"
+        var errMessageText = match[4]
```
Strauman: Sorry 😰

```suggestion
        let errMessageText = match[4]
```

lib/parsers/log-parser.js
Strauman: I'm not highly confident as to what ideally should a legal macro name in
this setting.

The problem is that the escape character, `\` by default, can be redefined, as well
as having anything being active letters (and thus treated as control sequences)

E.g. doing
```
\catcode`O=13\relax O
```
would make the letter O an unknown control sequence, and

```
\catcode`J=0\relax JUndefined
```
would make `JUnedfined` behave identically to `\Undefined` and this atom plugin
will report weird information.

You can't just look at the last space and go from there. If you consider
```
{}\some@undef@macro
```
Then the `{}` would be a part of the macro name.

However, I do argue that these cases are extraordinary circumstances,
and isn't really that necessary to implement.

lib/parsers/log-parser.js
```diff
@@ -76,10 +85,27 @@ export default class LogParser extends Parser {
...
       if (match) {
+        let outMessageText
         const lineNumber = match[2] ? parseInt(match[2], 10) : undefined
+        // Check if the error is "Undefined control sequence"
+        let errMessageText = match[4]
+        if (errMessageText === UNDEF_CONTROL_SEQ_TEXT) {
+          // The actual macro is the last thing
+          // in the sentence of the next line
+          let nextLine = lines[index + 1]
+          let macroMatch = nextLine.match(LAST_MACRO_PATTERN)
```
thomasjo: Let's declare `errMessageText`, `nextLine`, and `macroMatch` as `const`.

lib/parsers/log-parser.js
```diff
@@ -76,10 +85,27 @@ export default class LogParser extends Parser {
 
       match = line.match(ERROR_PATTERN)
       if (match) {
+        let outMessageText
```
thomasjo: I think maybe this variable should be named `outputMessage`. The `Text` suffix is a bit redundant.

lib/parsers/log-parser.js
```diff
@@ -76,10 +85,27 @@ export default class LogParser extends Parser {
...
         const lineNumber = match[2] ? parseInt(match[2], 10) : undefined
+        // Check if the error is "Undefined control sequence"
+        let errMessageText = match[4]
```
thomasjo: What do you think about renaming this to `errorMessage`?

lib/parsers/log-parser.js
```diff
@@ -76,10 +85,27 @@ export default class LogParser extends Parser {
...
+        if (errMessageText === UNDEF_CONTROL_SEQ_TEXT) {
+          // The actual macro is the last thing
+          // in the sentence of the next line
+          let nextLine = lines[index + 1]
```
thomasjo: I guess in most situations this will be safe given how the error is logged, but if we wanted to be super safe we should probably check that this index lookup is not out of bounds. One possibility is to extend the condition above like

``` js
if (errMessageText === UNDEF_CONTROL_SEQ_TEXT && lines.length > index) {
```

Although that might be wrong place to short-circuit this depending upon what error message you want to give to the user in that case. If you're pretty sure we'll never go out of bounds here, we can skip this check.

lib/parsers/log-parser.js
```diff
@@ -76,10 +85,27 @@ export default class LogParser extends Parser {
...
+          let macroMatch = nextLine.match(LAST_MACRO_PATTERN)
+          // Check if there is a match
+          if (macroMatch) {
+            outMessageText = errMessageText + ': ' + macroMatch[1]
```
thomasjo: Replace this with

``` js
outMessageText = `${errMessageText}: ${macroMatch[1]}`
```

spec/process-manager-spec.js
```diff
@@ -38,7 +38,7 @@ describe('ProcessManager', () => {
     processManager.killChildProcesses()
     processManager.executeChildProcess(constructCommand('new.tex'), { allowKill: true }).then(result => { newKilled = true })
 
-    waitsFor(() => oldKilled, 5000)
+    waitsFor(() => oldKilled, 10000)
```
thomasjo: Let's try removing these; I'm hoping whatever the underlying issue on AppVeyor was, that is has been resolved by now. Either way, your changes should not cause any errors in these tests, so if these are still timing out, that should be addressed separately.

thomasjo (changes requested):
I would really love it if we had a test covering this. The easiest way would probably be to use one of the log files in the fixtures folder (e.g. `file.log`) and augment it to contain an undefined control sequence error. Alternatively create a file which produces that error, and alter the locally generated log file by censoring local file paths, etc, like we've done for existing fixtures.

Having a test covering at least the "happy path" of this new logic ensures correct behavior, guards against regressions, and what not. Having tests that cover cases where a macro is not found would be amazing, but not strictly required.

lib/parsers/log-parser.js
Strauman: Agreed!

lib/parsers/log-parser.js
Strauman: Yes. I am a fan of explicit names. `errorMessage` it is!

lib/parsers/log-parser.js
Strauman: I can't think of a scenario where the the error would be `Undefined control sequence` and no control sequence will follow in the line below. However, I'll do an out-of-bounds check, to be sure?

lib/parsers/log-parser.js
Strauman: Whoop! Didn't know this was possible 🥇 🔢

Strauman:
> Having tests that cover cases where a macro is not found would be amazing, but not strictly required.

I'll see if I get the time.

spec/process-manager-spec.js
Strauman: Done!

Issue #480, opened 2018-08-09
Latexmk fails to work due to bad filename(s)


GauSyu:
<img width="994" alt="2018-08-09 15 59 17" src="https://user-images.githubusercontent.com/11472934/43887096-5a2a0e7e-9bf0-11e8-9964-caf69bc6d6f4.png">
As the log showing, latexmk fails to work because of my filename. However, this happens only after I update my Texlive: it works fine in the past, for the same file. I wonder what causes this problem and how to fix it (probably through relative path, which i don't know how to set up).

dsferruzza:
Same here! (with MiKTeK on Windows 10)

I'm pretty sure this is because I updated latexmk from `4.55` to `4.59`.
I'm still able to compile my project using `latexmk -pdf article.tex` from the command line, only atom-latex gives me this error.

CarlosEvo:
I can confirm the issue. I am using `MacTeX` on macOS 10.12. It happens in `atom-latex`, but not in terminal.

``` zsh
$ latexmk -version

Latexmk, John Collins, 7 August 2018. Version 4.59
```

Aerijo:
@thomasjo This just happened to me; I was also working an iCloud folder, and had just updated everything with `tlmgr`. Downgrading back to the stored backup version of `latexmk` made the issue go away, so it's definitely related to that recent(?) update.

I couldn't reproduce in a folder within my home directory, so I'm guessing the iCloud path is important (most likely the space in `Mobile Documents`).

Aerijo:
`latexmk` [deliberately throws an error](https://ctan.org/ctan-ann/id/mailman.6785.1533755937.5100.ctan-ann@ctan.org) now when the provided command line path has invalid(?) characters. I say invalid(?), because it was working fine for years in iCloud before this arbitrary change.

He doesn't seem to have added a flag to disable this either. I guess the only way to get around it would be to use relative paths. I believe that's why it still worked fine directly in the terminal.

polestarneighbor:
I've run into this issue as well, with MikTeX and Windows 10, on a filename that does not include tildes (but does include backslashes, because Windows).

scrouthtv:
Same for me (Windows 10 / texlive). Anyone knows how to fix?

thomasjo:
[...]

PR #499, opened 2018-11-26
Add relative path support

Aerijo:
(tries to) Fixes https://github.com/thomasjo/atom-latex/issues/480

Honestly, I have only tried this with a single file and no included / inputed / other fancy projects. That's why I want to leave the option disabled by default, but there if a user experiences the same issue as above.

I will report back with more attempts to break it, but I don't think I can test all combinations of files this package supports (sage, knitr, etc.). It doesn't help that these tests are failing anyway for me (I suppose it's because I don't have those programs installed? IDK).

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,18 @@ some important features.
 
 Any and all help is greatly appreciated!
 
+**NOTE:** `latexmk` does not support file paths containing special characters such as `~`. To partially circumvent this, add `useRelativePaths: true` to your config file like so
+```cson
+# config.cson
+"*":
+  "atom-latex":
+    useRelativePaths: true
+```
+When set, this package will use a relative path in place of an absolute one. This will allow `latexmk` to compile projects stored in directories that contain special characters. Note that the project itself must not contain special characters in its directory or file names.
+
+This feature has not been fully tested yet, and there are no guarantees it will work in all cases. Please raise an issue if you find a case where it fails.
+
+
 <!--refs-->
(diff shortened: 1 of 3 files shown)


Aerijo:
OK, can confirm it works for `input` files, even when it's all in a directory path containing `~`. I'd suggest merging this, and handling any further errors users report (as the change is entirely reversible and only enables builds that would have entirely failed otherwise).

thomasjo:
On first glance, this looks solid! I'll try to get around to testing and merging this during the weekend.

thomasjo:
Relative execution can be achieved by using the `-cd` option for `latexmk`. I'm also not convinced that we should introduce an option for this. I have a long-term goal of reducing the number of options we currently have before (finally) cutting a stable v1 release. 

If we cannot work around the file path validation in `latexmk` outside of using relative execution, then we should always use that execution type. That would mean we have to reconsider whether we want to have all output located in the root project directory or not. Perhaps having it located relative to the processed file is fine, perhaps even better than the current behavior?

@Aerijo What are you thoughts?

README.md
```diff
@@ -137,6 +137,18 @@ some important features.
 
 Any and all help is greatly appreciated!
 
+**NOTE:** `latexmk` does not support file paths containing special characters such as `~`. To partially circumvent this, add `useRelativePaths: true` to your config file like so
+```cson
+# config.cson
+"*":
+  "atom-latex":
```
thomasjo: This should be `latex`

Aerijo: I had a rogue `atom-latex` entry in my config, so I assumed it was like that. Fixed.

thomasjo:
Thanks! :sparkling_heart:

PR #500, opened 2018-11-30
Update dbus to remove flatmap vulnerability

Aerijo:
@thomasjo
See https://discuss.atom.io/t/the-event-stream-incident-and-the-atom-teams-response/60807/3 and  https://github.com/sidorares/dbus-native/issues/253


thomasjo:
Thanks for this! I've just published 0.50.1 that contains this fix :bowing_man:

